PR #2, opened 2015-11-03
Added default variable handling.

Marie89:
I added a JSON file (agile2json.py output) in data/ to read in an istream and I know there is a way of saving the file from the agile2json.py output via the makefile, but due to time constraints I'm not looking into this now.

### DIFF
--- a/converters/agile2json.py
+++ b/converters/agile2json.py
@@ -15,7 +15,8 @@ def _run():
     out_dict = {
         'layers': _get_layers(network),
         'inputs': _get_inputs(network),
-        'outputs': _get_outputs(network)
+        'outputs': _get_outputs(network),
+        'defaults': _get_defaults(network)
         }
     print(json.dumps(out_dict, indent=2))
 
@@ -55,6 +56,18 @@ def _get_inputs(network):
 
 def _get_outputs(network):
     return network['target_order']
+
+def _get_defaults(network):
+    """                                                                                                                                                        
+    Get the input defaults from AGILEPack.                                                                                                                     
+    """
+    defaults = []
+    for input_name in network['input_order']:
+        if input_name in network['default_values']:
+            default_val = network['default_values'][input_name]
+            #defaults.append({input_name: default_val})
+            defaults.append({'name': input_name, 'default_value': default_val})
+    return defaults
 
 if __name__ == '__main__':
     _run()
--- a/src/lwtag-test-full_default.cxx
+++ b/src/lwtag-test-full_default.cxx
@@ -1,0 +1,62 @@
+#include "LightweightNeuralNetwork.hh"
+#include "parse_json.hh"
+
+#include <Eigen/Dense>
+
+#include <iostream>
+#include <string>
+#include <sstream>
+#include <fstream>
+
+int main(int argc, char* argv[]) {
+  // Read in the configurationusing an istringstream.
+  std::ifstream weight_file("/Users/marie/Work/GitHub/lw-client/data/test-nn.json");
+  std::string line_str;
+  std::string weight_file_str;
+  while (std::getline(weight_file, line_str))
+    {
+      weight_file_str += line_str+ ' ' ;
+    }
+  std::istringstream weight_file_sstream(weight_file_str);
+  lwt::JSONConfig config = lwt::parse_json(weight_file_sstream);
+  std::map<std::string,double> m_map_defaults = lwt::get_defaults_from_json(weight_file_sstream);
+
+  // initialize the tagger from the configuration
+  lwt::LightweightNeuralNetwork tagger(config.inputs, config.layers, config.outputs);
+
+  // build some dummy inputs and feed them to the tagger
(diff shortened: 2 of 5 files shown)


src/lwtag-test-full_default.cxx
```diff
@@ -0,0 +1,62 @@
...
+#include <Eigen/Dense>
+
+#include <iostream>
+#include <string>
+#include <sstream>
+#include <fstream>
+
+int main(int argc, char* argv[]) {
+  // Read in the configurationusing an istringstream.
+  std::ifstream weight_file("/Users/marie/Work/GitHub/lw-client/data/test-nn.json");
```
dguest: make this path relative, or just use `argv[1]`

converters/agile2json.py
```diff
@@ -56,5 +57,17 @@ def _get_inputs(network):
 def _get_outputs(network):
     return network['target_order']
 
+def _get_defaults(network):
+    """                                                                                                                                                        
+    Get the input defaults from AGILEPack.                                                                                                                     
+    """
+    defaults = []
```
dguest: why not store the defaults as a dict (rather than a list with two fields?)

converters/agile2json.py
```diff
@@ -56,5 +57,17 @@ def _get_inputs(network):
...
+    for input_name in network['input_order']:
+        if input_name in network['default_values']:
```
dguest: You could just loop over `network['default_values']` here, rather than looping over 'input_order' and checking 'default_values'

converters/agile2json.py
```diff
@@ -15,7 +15,8 @@ def _run():
     out_dict = {
         'layers': _get_layers(network),
         'inputs': _get_inputs(network),
-        'outputs': _get_outputs(network)
+        'outputs': _get_outputs(network),
+        'defaults': _get_defaults(network)
```
dguest: AGILEPack doesn't add defaults by, um… default, so it may be worth checking to see if they exist before calling this function

src/lwtag-test-full_default.cxx
```diff
@@ -0,0 +1,62 @@
...
+
+  // build some dummy inputs and feed them to the tagger
+  lwt::LightweightNeuralNetwork::ValueMap input{
+    {"in1", 1}, {"in2", 2}, {"in3", NAN}, {"in4", 4} };
+  
+  std::map<std::string,double> corrected_input = input; // create copy of input map to ensure nothing is missing
+
+  // check inputs and replace defaults           
+  for (const auto in:input) {
+    for (const auto in_default:m_map_defaults) {
```
dguest: You shouldn't have to loop over a map and then check it's key against the value you're looking for. Just use something like `if (m_map_defaults.count(in.first)) { ... }`

src/lwtag-test-full_default.cxx
```diff
@@ -0,0 +1,62 @@
...
+  std::ifstream weight_file("/Users/marie/Work/GitHub/lw-client/data/test-nn.json");
+  std::string line_str;
+  std::string weight_file_str;
+  while (std::getline(weight_file, line_str))
+    {
+      weight_file_str += line_str+ ' ' ;
+    }
+  std::istringstream weight_file_sstream(weight_file_str);
+  lwt::JSONConfig config = lwt::parse_json(weight_file_sstream);
+  std::map<std::string,double> m_map_defaults = lwt::get_defaults_from_json(weight_file_sstream);
```
dguest: the `m_` prefix is for member variables…

src/parse_json.cxx
```diff
@@ -48,6 +48,26 @@ namespace lwt {
...
+    std::map<std::string,double> default_map;
+    for (const auto& d: pt.get_child("defaults"))
+      {
+	std::string name = d.second.get<std::string>("name");
+	auto def_val = d.second.get<double>("default_value");
+        default_map.emplace(name,def_val);
+      }
+    for(const auto& a: default_map)
+      {
+	std::cout << "default_map: " << a.first << "  " << a.second << std::endl;
```
dguest: don't print anything unless you're debugging

dguest:
General note: the json file is useful to make sure the C++ code works, but it would be better to include the yaml file, since we also need to be able to test the yaml -> json converter.

dguest:
I did some testing on my own, and it turns out that there are a few more subtleties to handling both NaN values and `std::istream` than I'd anticipated.

For one thing, the rectified layer had the bad habit of removing NaN values and replacing them with zero (since `NaN > 0` returns false), whereas I think it makes more sense that NaN propagates through every layer.

As for reading json from a `std::istream`, I realized that it doesn't really make sense to run `seekg` on some kinds of `std::istream` objects, for example on `std::cin`. Still, the `istream` interface is sort of nice (for reading from files, `std::cin`, etc) so I just moved all the json stuff into one function.

There's now an example called `lwtag-test-defaults` which tests all these additions. I'm going to close this pull request since the issues have now been addressed.

PR #6, opened 2016-03-24
starting dev of dense and highway

mickypaganini:
Added a dense and highway layer, and a quick test for them. However, this is not yet supported by the keras2json parser, so this is mostly just to propose a slightly different level of abstraction on the notion of a layer.

### DIFF
--- a/include/LightweightNeuralNetwork.hh
+++ b/include/LightweightNeuralNetwork.hh
@@ -17,6 +17,7 @@
 #include <vector>
 #include <stdexcept>
 #include <map>
+#include <utility>
 
 namespace lwt {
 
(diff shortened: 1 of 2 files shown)


dguest:
Hmm, honestly I'm not super keen on the dense layer. The main problem is that templating means we have to know something about the architecture at compile time, which I'd rather avoid for code that goes into our software releases (e.g. anything in Athena). Splitting the configuration also opens the door to a lot of bugs with mismatched training and architecture. 

But if you just want to combine a matrix layer, a bias layer, and an activation function, you should be able to use `Stack dense_layer(n_inputs, {layer_config});`, since the `Stack` class is just a way to combine an arbitrary number of layers.

Either way, I'm really excited to see highway layers implemented. If I understand correctly, the implementation should be very similar to LSTM nets, so maybe they can be written to use some of the same classes.

dguest:
superseded by #18

PR #7, opened 2016-04-20
adding LightweightRNN

makagan:
merging Lightweight RNN, which contains recurrent layers, back into head


makagan:
sorry, need to add something before merging...

PR #9, opened 2016-04-20
added keras2json converter for Keras version 1.0.0

Marie89:
I adapted the keras2json converter with respect to the output conventions used in Keras version 1.0.0.

In addition, the version number gets saved in the input HDF5 file and a warning is printed in case the Keras output files were generated using another Keras version.


dguest:
Awesome! If I can confirm that everyone is using Keras 1.0 or later I can just replace the current converter with this one.

dguest:
I merged the commit where you added this into my master branch.

PR #18, opened 2016-05-07
Add Dense and Highway

mickypaganini:
Hey, I tried adding Dense and Highway layers again. This is very similar to my very first pull request, except that I adapted my layers to the new way of using activation functions. I added a short test-highway.cxx file. Everything is making correctly and running fine. I haven't checked yet whether the layers are working as expected and the numerical values make sense.
I hope this is what you wanted. Feel free to change it around if that's not exactly what you wanted me to do.


dguest:
This is great! I may rewrite some of the internals a bit to flatten the structure (I'm trying to avoid having too many nested classes), but if we can confirm that this does as expected it's already an awesome addition. Thanks!

mickypaganini:
Sounds good! I checked the performance using Keras and fixed a bug I had -- thanks for the suggestion. Check out the new pull request I sent you (https://github.com/dguest/lwtnn/pull/19)

PR #20, opened 2016-05-08
GRU

mickypaganini:
I implemented GRUs by following the example of LSTMs. Some things might be redundant and could probably be simplified and cleaned up. 

I tested it using `lwtnn-test-rnn.cxx` and that gave me the following output, which matches what I get with Keras, as you can see from this gist: http://nbviewer.jupyter.org/gist/mickypaganini/a9a57e67c130cf77c11605938c6cb5a1

```
[mp744:~/Documents/CERN/Golling/lwtnn] rnn(+11/-10)* 6s ± ./bin/lwtnn-test-rnn nn_config.json 
running over 1 loops
running fast
 -0.999984  -0.866316
 -0.736924  -0.165028
  0.511211   0.373545
-0.0826997   0.177953
 0.0655345   0.860873
 -0.562082   0.692334
 -0.905911  0.0538576
  0.357729   -0.81607
  0.358593   0.307838
  0.869386  -0.168001
 -0.232996   0.402381
 0.0388327   0.820642
  0.661931   0.524396
 -0.930856  -0.475094
 -0.893077  -0.905071
 0.0594004   0.472164
  0.342299  -0.343532
 -0.984604   0.265277
 -0.233169   0.512821
output sum:
0.285286
0.329717
0.161822
0.223176
```

`nn_config.json` was obtained by running `python converters/keras2json.py ../IPNN/RNN.json ../IPNN/variables.json ../IPNN/weights/ip3d-replacement_RNN.h5 > nn_config.json`

`variables.json` is produced here (https://github.com/mickypaganini/IPNN/blob/master/dataprocessing.py#L130) according to the model provided in `keras2json.py`

### DIFF
--- a/converters/keras2json.py
+++ b/converters/keras2json.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 #
 # Converter from Keras (version 1.0.0) saved NN to JSON
 """
@@ -185,7 +185,7 @@ def _gru_parameters(h5, layer_config, n_in):
             'bias': layers['b_' + gate].flatten().tolist(),
         }
         # TODO: add activation function for some of these gates
-    return {'gru_components': submap, 'architecture': 'lstm',
+    return {'gru_components': submap, 'architecture': 'gru',
             'activation': layer_config['activation'],
             'inner_activation': layer_config['inner_activation']}, n_out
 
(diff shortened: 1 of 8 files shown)


dguest:
This looks awesome. 

The only minor issue is that I'm not so sure about adding a new component class for each unit. We may be able to simplify code in the future if we just stick to one `Components` `enum class` and one map in the `LayerConfig` that holds them.

The same holds for the highway layers: it's probably easiest to store each dense unit in the `components` map within the `LayerConfig` class. Since each layer will only ever have one set of components, I think it's safe to merge them.

PR #22, opened 2016-05-10
Full Highway Support

mickypaganini:
Highway layers are now fully supported from `keras2json.py` to prediction. 
I use `bin/lwtnn-test-arbitrary-net` to make sure my output is exactly the same as that obtained using Keras (shown here: https://gist.github.com/mickypaganini/2454d56a836cac537f5265945f10f575)


converters/keras2json.py
```diff
@@ -126,7 +126,34 @@ def _get_dense_layer_parameters(h5, layer_config, n_in):
...
+        'activation': _activation_map[layer_config['activation']],
+    }
+    return return_dict, weights.shape[1]
+
+def _get_highway_layer_parameters(h5, layer_config, n_in):
+    """Get weights, bias, and n-outputs for a highway layer"""
+    layer_group = h5[layer_config['name']]
+    layers = _get_h5_layers(layer_group)
+    weights = layers['W']
+    weights_carry = layers['W_carry']
```
dguest: we could store the components here the same way we do it in the LSTM and GRU layers, we don't need to add two extra special cases

converters/keras2json.py
```diff
@@ -275,9 +303,18 @@ def _parse_inputs(keras_dict):
...
+    if len(keras_dict['class_labels']) == 2:
+        classes = [keras_dict['class_labels'][1]]
+    else:
+        classes = keras_dict['class_labels']
+
     out = {
         'inputs': inputs,
-        'outputs': keras_dict['class_labels'],
+        'outputs': classes,
         'defaults': defaults,
```
dguest: This seems like it introduces more problems than it solves. What if someone wants a nn for regression, and wants (x,y) outputs, for example? We don't want to throw away y…

include/lwtnn/NNLayerConfig.hh
```diff
@@ -33,6 +33,8 @@ namespace lwt {
     // dense layer info
     std::vector<double> weights;
     std::vector<double> bias;
+    std::vector<double> weights_carry; // for Highway
```
dguest: if we just treat the "carry" and "non-carry" dense matrices as components like we do with LSTM or GRU, we can avoid adding these two vectors, and that lets us make the C++ json parsing more general.

src/LightweightNeuralNetwork.cxx
```diff
@@ -160,13 +199,80 @@ namespace lwt {
...
+    }
+
+    // add activation layer
+    Activation a = Activation::LINEAR;
+    if (layer.activation != Activation::LINEAR) {
+      a = layer.activation;
+    }
+
+    _layers.push_back(new HighwayLayer(W, b, W_carry, b_carry, a));
     return n_outputs;
```
dguest: I think the above block could be condensed considerably if we stored the layer by component

src/lightweight_nn_streamers.cxx
```diff
@@ -64,26 +65,37 @@ std::ostream& operator<<(std::ostream& out, const lwt::LayerConfig& cfg){
...
+
   out << "weights: [" << cfg.weights << "]\n";
   out << "bias: [" << cfg.bias << "]\n";
+
+  if (cfg.weights_carry.size() > 0) { // for highway
+    out << "weights_carry: [" << cfg.weights_carry << "]\n";
+    out << "bias_carry: [" << cfg.bias_carry << "]\n";
+  }
+
   if (cfg.U.size() > 0) {
```
dguest: same comment on using components

src/lwtnn-test-arbitrary-net.cxx
```diff
@@ -60,14 +60,27 @@ namespace {
...
     }
+    
+    // for (const auto& input: config.inputs) {
+    //   in_vals[input.name] = -input.offset;
+    // }
     auto out_vals = tagger.compute(in_vals);
     for (const auto& out: out_vals) {
-      std::cout << out.first << " " << out.second << std::endl;
+      std::cout << "Class: " << out.first << std::endl;
+      std::cout << "Probability = " << out.second << std::endl;
```
dguest: Unfortunately this breaks the unit test I had before. Having these listed on two lines is nice for human readability, but it makes it a lot harder to write a test around it

src/lwtnn-test-highway.cxx
```diff
@@ -0,0 +1,44 @@
...
+
+	// -- Test Highway:
+	lwt::HighwayLayer hw(W, b, 0.17 * W, -3 * b, lwt::Activation::RECTIFIED);
+	std::cout << "Predictions from lwtnn Highway:" << std::endl;
+	std::cout << hw.compute(x) << std::endl;
+}	
+
+// These predictions are to be compared to the ones obtained with Keras and found here:
+// http://nbviewer.jupyter.org/gist/mickypaganini/e997d21682a24e6e8c474af25760afa0
+
```
dguest: this is cool, I like standalone checks

src/parse_json.cxx
```diff
@@ -14,6 +14,7 @@ namespace {
   lwt::Architecture get_architecture(const std::string&);
   void set_defaults(LayerConfig& lc);
   void add_dense_info(LayerConfig& lc, const ptree::value_type& pt);
+  void add_highway_info(LayerConfig& lc, const ptree::value_type& pt);
```
dguest: as I've commented above, I think we could basically merge this with the LSTM and GRU parser

dguest:
I just patched a bunch of changes from this branch into my `merge-highway` branch. Unfortunately I can't check to make sure it works because I don't have the trained network! Any way you can send me that? (or you can check it, but since I made all these changes maybe it's on me)

dguest:
Hi, I had to fix a few bugs after my rewrite, but following those fixes my output looks identical to the one you had with the files you sent. I've added a test around those files, thanks!

PR #29, opened 2016-11-25
Normalization layer added to converter

jwsmithers:
It seems only the scaling and shifting outputs of the normalization layer is needed
when applying this in a trained NN. Therefore, weights = gamma, and
bias = beta. running mean and std may not be needed.
https://arxiv.org/pdf/1502.03167v3.pdf

### DIFF
--- a/include/lwtnn/Stack.hh
+++ b/include/lwtnn/Stack.hh
@@ -69,6 +69,7 @@ namespace lwt {
     // returns the size of the next layer
     size_t add_layers(size_t n_inputs, const LayerConfig&);
     size_t add_dense_layers(size_t n_inputs, const LayerConfig&);
+    size_t add_normalization_layers(size_t n_inputs, const LayerConfig&);
     size_t add_highway_layers(size_t n_inputs, const LayerConfig&);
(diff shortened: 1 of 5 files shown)


dguest:
Closing and reopening to trigger CI

include/lwtnn/Stack.hh
```diff
@@ -136,6 +137,26 @@ namespace lwt {
...
+
+  public:
+    NormalizationLayer(const MatrixXd& W,const VectorXd& b);
+    virtual VectorXd compute(const VectorXd&) const;
+
+  private:
+    MatrixXd _W;
+    VectorXd _b;
+
+    int _n_outputs;
```
dguest: This doesn't seem to be used (same for `_return_sequences` below)

include/lwtnn/Stack.hh
```diff
@@ -136,6 +137,26 @@ namespace lwt {
...
+
+    bool _return_sequences;
```
dguest: As I said above, this doesn't seem to be used. Otherwise everything looks ok to me.

Issue #28, opened 2016-11-01
Add normalization layer


dguest:
Having a "normalization" layer could be useful in a few ways:
 - We could replace the `InputPreprocessor` with a normalization layer
 - More importantly, we could support "batch" normalization without the batches

Keras stores the running mean and standard deviation from batch normalization, so converting this into layer information should be relatively easy.

Marie89:
Given the improvements we see using it for b-tagging (see [FTag Algo meeting update on DL1](https://indico.cern.ch/event/571869/#preview:2091811)), it would be good to add it to rel21 while it's still possible. Meaning within the next week. 
Is this timescale feasible?

jwsmithers:
Hi Marie, 
This can potentially happen by, say, Wednesday? Would that be soon enough?

Marie89:
Hey Josh,
if this means that it's validated on a simple architecture, implemented in rel21 and the [Keras2JSON converter](https://github.com/lwtnn/lwtnn/blob/master/converters/keras2json.py) is adapted correctly by say Wednesday noon, I guess yes. FYI, I'm currently using Keras version 1.1.1 and w.r.t. BN there were a few changes in the previous versions, so in case people are using older versions and would like to benefit from using it in Athena. I can run a last validation check using a NN that uses BN on Wednesday afternoon if it's working for DL1 - just as a very last validation check.
But just to make sure, I'd also ask Richard and Manu on the timeline they have in mind for closing the release to changes.

jwsmithers:
The converter is done, it's just implementing the layer now and validating which I suspect will be the hardest part. I'll probably be asking Dan for help.
I'll shoot for having this done by Wednesday, but this will be very tight.

dguest:
This should be complete as of pull request #30

PR #30, opened 2016-11-29
(Working) Normalization layer added

jwsmithers:
Bulk of normalization done in converter. Stack.cxx does a scale and shift operation.

### DIFF
--- a/src/Stack.cxx
+++ b/src/Stack.cxx
@@ -1,5 +1,6 @@
 #include "lwtnn/Stack.hh"
 #include <Eigen/Dense>
+#include <iostream>
 
 #include <set>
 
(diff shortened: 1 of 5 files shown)


src/Stack.cxx
```diff
@@ -1,5 +1,6 @@
 #include "lwtnn/Stack.hh"
 #include <Eigen/Dense>
+#include <iostream>
```
dguest: `iostream` is nice for debugging but I'd prefer to remove it from here. Also getting rid of it is a good way to ensure that there aren't any `cout` statements lurking.

src/Stack.cxx
```diff
@@ -207,6 +231,18 @@ namespace lwt {
...
 
+   // Normalization layer
+   NormalizationLayer::NormalizationLayer(const MatrixXd& W,
+                                          const VectorXd& b):
+    _W(W), _b(b)
+  {
+  }
+  VectorXd NormalizationLayer::compute(const VectorXd& in) const {
+    VectorXd shift = in + _b ;
+    MatrixXd _W_t = _W.transpose();
```
dguest: We're using the leading `_` to indicate that variables are owned by the object, so calling a local variables `_W_t` is a bit confusing. But more importantly, although this may be optimized out somehow, I don't think we want the transpose in `compute(...)`: this gets called every time we classify something, whereas if we just store the transposed version it gets called once in the constructor.

dguest:
Looks great! This closes issue #28

PR #32, opened 2016-12-02
get epsilon value from json file for normalization layer.

jwsmithers:
This assumes that the value of epsilon does not change from 1 BN layer to the next (i.e. someone manually changing the value in each BN layer they create). Since this value is really just a fudge factor that people shouldn't be setting anyway, I doubt this will happen.

### DIFF
--- a/converters/keras2json.py
+++ b/converters/keras2json.py
@@ -144,7 +144,7 @@ def _normalization_parameters(h5, layer_config, n_in):
     assert mean.shape[0] == stddev.shape[0]
     assert gamma.shape[0] == n_in
     assert 'activation' not in layer_config
-    epsilon = 1e-05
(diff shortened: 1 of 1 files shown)


dguest:
Even if the epsilon value changes between layers, we have a different configuration dictionary in the json for each layer, right? So I think this should be fine. I'm merging it but please let me know if you think that's really a concern.

dguest:
Also in the future try to merge with the master branch in this repository before you commit. I fixed some links in the README that you haven't merged, which prevented the last merge from being a simple fast forward.

jwsmithers:
Actually, I'm being dumb. This epsilon layer will work if changed between layers... I just tested it.
It's grabbing the value for each layer.

jwsmithers:
Ah ok, will do!

Issue #35, opened 2017-02-10
building lwtnn with custom boost


smuzaffar:
In order to use our boost installation we had to patch makefile 
https://github.com/smuzaffar/cmsdist/blob/cc07391e4f1b8f6a4b00ce17fbc604ada7b539a0/lwtnn-1.0-boost-fix.patch

would it be possible to provide a way so that custom boost and flags can be set instead of patching the makefile?

dguest:
Would this solve your problem: https://github.com/lwtnn/lwtnn/commit/7d5a8fcc47a985eba68b5abe935a0f43b58dc3af
?

smuzaffar:
thanks, this should be enough.

dguest:
OK, merged into master

PR #36, opened 2017-02-12
Allow the user to set compile options

dguest:
The makefile now checks for three environment variables:

 - LWTNN_CXX_FLAGS: extra compiler options
 - LWTNN_LDFLAGS: extra linker flags
 - LWTNN_LIBS: extra libraries

Hopefully this makes it easier to avoid hacking the makefile in custom
environments.


dguest:
This should close #35

PR #44, opened 2017-04-17
Keras2 support

jwsmithers:
Changes to support Keras 2.0. Mainly support for the functional API converter.
For the sequential model, the fallback is Keras 1.2 and backend=theano.


converters/kerasfunc2json.py
```diff
@@ -217,6 +230,7 @@ def _build_layer(output_layers, node_key, h5, node_dict, layer_dict):
 
     layer_type = node.layer_type
     if layer_type in skip_layers:
+        node.n_outputs = sum(s.n_outputs for s in node.sources)
```
jwsmithers: This causes issues later on, when our node count includes these skipped layers.

dguest:
I took a look at this. It looks like supporting keras 2 is going to require changes in pretty much every part of the layer converters. If it saves you some trouble, I'd say don't worry about backward compatibility: we could just focus on getting something that works for Keras 2, and then create two versions of the converter and migrate whatever common functions we can into a module. 

Do you think this would make this easier? I'm just worried that having if statements for every function which checks the Keras version is going to get ugly. 

The changes to support two back ends, on the other hand, look pretty manageable and we could probably handle that with if statements or something similar.

jwsmithers:
Well, actually the thing that's taking me A LOT of time to figure out is how to skip nodes when the parent is a dropout (or any other layer we don't want to rebuild). For functional API i'm just trying to get this working and not really focus on back-porting this to keras v1 (since functional isn't as important here). We spoke about this the other day, but tbh, I'm still not sure how to actually do it.

In terms of v1 vs v2 for actual layers: I've so far just done the dense layer which requires 6 extra lines to make it compatible with tensorflow, theano, keras v1 and v2, (excluding the extra _send_recieve_meta infomation). I image the other layers will be similar. So it will add a few more if statements, but from the user's (and our) point of view it might be easier to just maintain one converter script? 
Also, I can clean this up and just have one function that defines layers names to reduce if statements.

converters/keras_v2_layer_converters.py
```diff
@@ -0,0 +1,173 @@
...
+
+# TODO: unify LSTM, and GRU here, they do almost the same thing
+def _lstm_parameters(h5, layer_config, n_in):
+    """LSTM parameter converter"""
+    layer_group = h5[layer_config['name']]
+    layers = _get_h5_layers(layer_group)
+    n_out = layers['recurrent_kernel'+ BACKEND_SUFFIX].shape[0]
+    submap = {}
+    
+    for n_gate, gate in enumerate('ifco'):
```
jwsmithers: See https://github.com/fchollet/keras/blob/master/keras/layers/recurrent.py#L1018 for how it's split

converters/keras_v2_layer_converters.py
```diff
@@ -58,15 +59,24 @@ def _normalization_parameters(h5, layer_config, n_in):
     }
     return return_dict, scale.shape[0]
 
-# TODO: unify LSTM, and GRU here, they do almost the same thing
-def _lstm_parameters(h5, layer_config, n_in):
-    """LSTM parameter converter"""
+def _rnn_parameters(h5, layer_config, n_in):
+    """RNN parameter converter. We support lstm and GRU """
     layer_group = h5[layer_config['name']]
+    if "lstm" in layer_config['name']:
```
jwsmithers: hmm actually this isn't foolproof.
If someone renames their layer then this fails. I don't actually know how we can get around this except for adding two separate functions.

dguest: There should be an entry called "class_name" somewhere in the `layer_config` dictionary. It's the same one we use to find which layer converter to call.

jwsmithers: At this point there isn't. "layer_type" is passed into the layer file and used in layer_converters to create the dictionary which calls the respective function.

converters/keras_v2_layer_converters.py
```diff
@@ -18,7 +18,7 @@ def _send_recieve_meta_info(backend):
     global BACKEND_SUFFIX
     BACKEND_SUFFIX = ":0" if backend == "tensorflow" else ""
 
-def _get_dense_layer_parameters(h5, layer_config, n_in):
+def _get_dense_layer_parameters(h5, layer_config, n_in, *args):
```
jwsmithers: I could also just call it "layer_type" which would probably be better. I think I had something else in mind when I did this and then forgot what is was.

PR #53, opened 2017-07-26
CMake build

jwsmithers:
This is to get a handle on how to compile from scratch when I put in Athena and have to use cmake.
It's probably not a bad idea to have generally, but I'll leave it up to you whether or not you want this.

Something quite interesting, there are a loooot of warnings on lxplus (stupid indentation warnings etc. because of Eigen.), but none when building on Mac. Maybe we don't wan't -Wall flag set? not sure why it only displays them with cmake build and not normal make build.

Anyways, some instructions:

**Setting up the correct env in lxplus**
```
lsetup root 
localSetupSFT releases/LCG_88/eigen/3.2.9
localSetupBoost  boost-1.60.0-python2.7-x86_64-slc6-gcc49
export EIGEN_INC=/cvmfs/sft.cern.ch/lcg/releases/LCG_88/eigen/3.2.9/x86_64-slc6-gcc49-opt/include/eigen3
export BOOST_INC=export BOOST_INC=/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/boost/boost-1.60.0-python2.7-x86_64-slc6-gcc49/boost-1.60.0-python2.7-x86_64-slc6-gcc49/include
lsetup "cmake 3.6.0"
```
Setting up Root helps get cmtconfig tag and just helps with matching eigen automatically.
Paths need to be set explicitly for some reason. And you need a newer version of cmake.

**External to lxplus**
Just have a newer cmake, and point EIGEN_INC and BOOST_INC to respective folders (will complain and fail if can't find it.)
On my mac it's something like `/usr/local/include`.

**Compiling**
```
checkout lwtnn
mkdir build install; cd build
cmake -DCMAKE_INSTALL_PREFIX=../install ../lwtnn
make -j4
make install
```
Then add the headers to libraries to appropriate paths.

Give it a try, should work

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,8 @@ src/lwtnn-test-graph.cxx
 src/lwtnn-test-rnn.cxx
 )
 
+# Build each exectuable and link lwtnn
+# Also set it's install path when make install is called
 foreach( exec_full ${EXECUTABLES} )
     # I used a simple string replace, to cut off .cpp.
     string( REPLACE "src/" ""  execname_temp ${exec_full} )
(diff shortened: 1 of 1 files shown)


dguest:
Maybe warnings may have as much to do with the `-pedantic` flag as the `-Wall` flag? Anyway, I don't have a very good lxplus connection from here but I'll try it in the next few days.

PR #55, opened 2017-07-27
Cmake tweaks

jwsmithers:
Change the way in which boost and eigen are found. Now by default they are looked for in /usr/local/include. If not found, it fails and set the paths manually.

I.e. to use this on lxplus, you don't even need to setup anything fancy, but rather just point it to the headers.

```
lsetup "gcc 620_x86_64_slc6"
lsetup "cmake 3.6.0"
cmake -DEIGEN_INC=/cvmfs/sft.cern.ch/lcg/releases/LCG_88/eigen/3.2.9/x86_64-slc6-gcc49-opt/include/eigen3 -DBOOST_INC=/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/boost/boost-1.60.0-python2.7-x86_64-slc6-gcc49/boost-1.60.0-python2.7-x86_64-slc6-gcc49/include ../lwtnn
```
 
This is needed so that I can make use of this CMakeLists.txt file for the Athena build, see here: https://gitlab.cern.ch/jwsmith/atlasexternals/tree/out_source_lwtnn

I also removed some warnings.


jwsmithers:
For the AtlasExternals build, this needs to be merged (obviously) and we need a tagged version of lwtnn with this CMakeLists.txt to use.

https://gitlab.cern.ch/jwsmith/atlasexternals/blob/out_source_lwtnn/External/lwtnn/CMakeLists.txt

dguest:
There are a lot of hardcoded paths in this approach. Is there some way to make it use cmake's `find_packages` function to get `EIGEN3_INCLUDE_DIR` and `BOOST_INCLUDE_DIR`?

jwsmithers:
I'm copying and pasting my messages from slack here (since you're offline there):

the reason I didn’t use find_package() in the CMakeLists.txt is that it requires boost/eigen to be installed beforehand. When building AtlasExternals the assumption is you have a very clean build machine without these on them. So instead I point the variables to their include dirs. This means that AtlasExternals uses the same CMakeLists as in lwtnn, but the Eigen and boost paths are set at compile time and need to be built first.

For example, Exerces and Expat are two packages that are required to be yum installed on any build machine. So Geant4 CMakeLists tries to find them. But stuff like ROOT/CLHEP needs to be set on the fly because it’s matrix dependent 
`   -DCLHEP_ROOT_DIR:PATH=${CMAKE_BINARY_DIR}/${ATLAS_PLATFORM}`

This means that you don't need to set any paths when building within atlas, and if you're building locally it will look in usr/include for the headers. If those don't exist then you have to set them with the -DEIGEN_INC and -DBOOST_INC options when configuring. This is no different from how we do it with the original `make`.

https://gitlab.cern.ch/jwsmith/atlasexternals/blob/out_source_lwtnn/External/Geant4/CMakeLists.txt

dguest:
OK, I guess we just need something that works with AtlasExternals to start off with. It would be nice in the longer run if we could use `find_package` to look for boost and eigen if they aren't specified, but I'm not very good at CMake so I guess this will have to wait.

PR #60, opened 2017-08-03
Use cmake to find packages

dguest:


dguest:
Merge in changes to CMakeList.txt

This merge includes some cleanup to the CMakeList:
 - Executables are linked to a static library (rather than the shared one)
 - Packages are found using CMake's own `find_package` function rather than environment variables or hardcoded paths.
 - Source files are found by globbing rather than listed explicitly

PR #61, opened 2017-08-03
CMake Configuration Upgrade (2017.08.03.)

krasznaa:
Pull requests may not be the way to work with this repository, but I thought this would be the simplest to "quickly" demonstrate how I would propose the project's configuration to look like.

This is a continuation/extension of the discussion happening on https://gitlab.cern.ch/atlas/atlasexternals/merge_requests/106.

In this setup Boost is found with CMake's own `FindBoost.cmake` code (https://cmake.org/cmake/help/v3.0/module/FindBoost.html), while Eigen is found with the simple module bundled with this project. To guide `FindBoost.cmake`, one needs to use the variables explained on the linked webpage. To guide `FindEigen.cmake`, one needs to use either the `CMAKE_PREFIX_PATH` variable, or the `EIGEN_ROOT_DIR` (cache) variable. I.e.:

```
cmake -DEIGEN_ROOT_DIR=/the/local/Eigen/dir ../lwtnn/
```


dguest:
Hi @krasznaa,

Thanks! Why aren't pull requests the right way to work with this repository? This looks quite nice to me (and it passed all the tests).

I'll probably merge this as is unless you can think of a reason not to, but in the meantime I had a few questions:

 - The only question I have is on the `FindEigen.cmake` file you've included. Eigen seems to come with a [`FindEigen3.cmake`][1] module, would it be better to use the same name?
 - What's the logic behind hard-coding all the source files rather than using some kind of grep expression?
 - Is there some way to set the git tag as the version number? I could imagine doing it with something like [execute_process][2], is there some reason not to do this?

But again thanks a lot for cleaning this up!

[1]: https://bitbucket.org/eigen/eigen/src/8d1ccfd9c5a0bf1a662bd0fe35858f4e527d2bac/cmake/FindEigen3.cmake?at=default&fileviewer=file-view-default
[2]: https://cmake.org/cmake/help/v2.8.9/cmake.html#command:execute_process

jwsmithers:
As soon as i'm not completely slammed by my analysis and due dates I'll contribute to this conversation again :)

krasznaa:
Hi Dan,

Some projects only "work" with people asking for direct access to the repository, and making their changes inside of a feature branch of the main repository. (This is how Gaudi operates in principle, ATLAS just forced the merge request model on them. 😛) Yet other projects only accept "outside help" in the form of diffs sent to some mailing list. (CMake...) I realised only when setting up this push request that I didn't check it if you guys had some contribution guide which would advise against push requests of this form...

**For Eigen:** I was looking for a find-module provided by Eigen, but since I didn't see one in its installation, I thought they didn't make one themselves. But you're right, they do provide one called `FindEigen3.cmake` inside their source code. (They just don't install it with the project.) I chose the `FindEigen.cmake` name since that's the one that we use in the ATLAS code. Nothing deeper behind it.

In principle it could be a good idea to just take `FindEigen3.cmake` made by the Eigen developers, and put that into this project. It won't make a huge difference, but I do like using other people's code when possible.

**For the source files:** The CMake recommendation is to always hardcode source file names. We had to break with this inside the ATLAS code since for CMT we declared source file names almost always with wildcards. So the CMT->CMake conversion script had a much easier time putting the same wildcards into the `CMakeLists.txt` files than to evaluate those wildcards while creating the CMake configuration.

The idea here is that if you explicitly list all source file names, then you can't get into a situation when just executing `make` (or whatever other build system) would not get the right results. You see, if you use wildcards in your configuration, when you add a new source file to your code, you have to either do a clean build, or at least touch the `CMakeLists.txt` file of the project. While if you explicitly list all source files, then when you add a new one, you have no choice but to "touch" the CMake configuration.

So all in all, as long as the number of source files is O(20) or less, I would not use globbing expressions.

**For the version number:** You have to be careful here unfortunately. CMake only recognises strings of the form `[0-9]+(\.[0-9]+)+` as a "version number". I had a lot of struggle with this already. 😦 Even something like "1.0.3b" is something that freaks `find_package` out. And if you use a version number that you can't refer to directly with `find_package`, then what's the point of the version number anyway?

Note that this is why in order to find LCG releases of the form "LCG_78root6" or "LCG_88Py3" our AtlasLCG code has to be used like this:

```cmake
set( LCG_VERSION_POSTFIX "Py3" )
find_package( LCG 88 REQUIRED EXACT )
```

Yepp, pretty awkward. So it's easier to use a version number of the form "X.Y.Z" that you update on "major releases". This is anyway a good practice to force you to make tags of the repository regularly.

I'll make some tests with `FindEigen3.cmake`, and will most probably update this request shortly.

Cheers,
       Attila

krasznaa:
Hi Dan,

Hmm... What copyright do you want to use for this code? Eigen's `FindEigen3.cmake` code uses the Simplified BSD one.

I have very little idea about licensing to be honest. I **think** it's okay to include their file as-is, but you may need to declare all of lwtnn to use the Simplified BSD license because of it. Of course the question of licensing is even relevant if we don't include **any** Eigen code inside this project directly, but "just" use it. I just assume that that's okay because we also use Eigen inside the ATLAS code, and we didn't make that code public. (People more knowledgeable than myself have never identified Eigen as a problem source for this. But let's not discuss here what the questionable parts are in that code... 😛)

So. Before I'd update my push request, maybe one of you should try to check what it means to include `FindEigen3.cmake` from their source code directly inside this repository.

Cheers,
       Attila

dguest:
Hi,

*On Contributing:* We do have [contributing guidelines][1], where we encourage pull requests (even to the guidelines themselves if you think something is unclear). That's also why we host it on github, so your request is definitely welcome!

*Licensing:* This code is [under the MIT licence][2]. I don't know if we'll be causing conflicts by grabbing a bit of BSD code and putting it here, but [one answer on stack exchange][3] seems to suggest it's fine if you leave the original licence in the beginning of the file.

I'm a bit confused as to the best practices around `Find*.cmake` files, though: my main nitpick with your file was that I assumed most other projects would call it `FindEigen3.cmake` given the precedent from the Eigen developers. If future versions of CMake include the Eigen find files, (or if future versions of Eigen install the find file by default) it's easier if we're using the same name convention. So I'd also be fine with just renaming your file `FindEigen3.cmake`, especially if we're worried about licensing issues.

That said, I really don't have a _strong_ opinion either way: even the file you have now is completely fine.


[1]: https://github.com/lwtnn/lwtnn/blob/master/CONTRIBUTING.md
[2]: https://github.com/lwtnn/lwtnn/blob/master/LICENCE
[3]: https://softwareengineering.stackexchange.com/questions/278126/is-it-legal-to-include-bsd-licensed-code-within-an-mit-licensed-project

krasznaa:
As I wrote before, I like the idea of using the `FindEigen3.cmake` file coming from Eigen itself. So I exchanged my file for that one.

I think you're right, we should be fine license-wise. I just wanted to stop for a moment to think about that...

dguest:
Ok, thanks Atilla! I've merged this in because I have another bug to fix on top.

PR #63, opened 2017-08-04
Teach the project how to build all its externals manually (2017.08.04.)

krasznaa:
Let me propose another change as well then off of the success of the previous pull request. 😉

While Boost itself is a pretty common external, a vanilla linux/macOS machine usually doesn't have it installed. So I taught the project how to download and "build" both Boost and Eigen by itself if the user asks for it.

For this I introduced two "CMake options". (Boolean cache variables.)
   - `BUILTIN_BOOST`: When set to `TRUE`, the project downloads Boost 1.64.0 itself, and uses the headers from that download for the build.
   - `BUILTIN_EIGEN`: When set to `TRUE`, the project download Eigen 3.2.9. Apart from building the code against these downloaded headers, it also installs these headers along with the rest of the project. (Since the lwtnn headers publicly depend on Eigen.)

As a result one can now build the code on a pretty bare system with:

```sh
cmake -DCMAKE_INSTALL_PREFIX=/some/path -DBUILTIN_BOOST=TRUE -DBUILTIN_EIGEN=TRUE ../lwtnn/
```

The design/naming here comes from ROOT. They have a bunch of similar options themselves.

https://root.cern.ch/building-root

The other small tweaks are not so super important. The code now just makes sure that the build already fails at configuration time if the detected compiler can't do C++11.

Cheers,
         Attila


dguest:
This looks great to me! It would be nice to include this clean build procedure in the travis build matrix, I'll look into that at some point.

PR #70, opened 2017-09-19
ELU Supprt

demarley:
Add (minimal) support for Exponential Linear Unit activation function.


converters/keras_v1_layer_converters.py
```diff
@@ -187,6 +187,11 @@ def _get_merge_layer_parameters(h5, layer_config, n_in, layer_type):
             'activation': 'linear'}, sum_outputs
 
 
+def _get_elu_activation_parameters(h5, layer_config, n_in):
+    """Return dummy parameters for ELU activation"""
+    return {'weights':[], 'bias':[], 'architecture':'dense','alpha':layer_config["alpha"],
+            'activation':_activation_map['elu']}, n_in
+
```
dguest: I'm not sure that this function ever gets called: are there any layers called "elu"?

converters/keras_v1_layer_converters.py
```diff
@@ -205,6 +210,7 @@ def _activation_parameters(h5, layer_config, n_in, layer_type):
     'merge': _get_merge_layer_parameters,
     'activation': _activation_parameters,
     'timedistributed': _time_distributed_parameters,
+    'elu': _get_elu_activation_parameters,
```
dguest: similar here: I think the `elu` is probably only ever seen by the activation function part of the code.

jwsmithers: So I've only ever seen `elu`'s called as i.e.

`add(Dense(1, activation='elu'))`.

Which means it's just an activation. But I haven't looked at the changes in detail yet.

demarley: I'm not sure I follow the discussion exactly, but here is how we implemented ELU:
https://gitlab.cern.ch/AllHadAC/CyMiniAnaAC/blob/master/python/deepLearning.py#L148

And we used an older converter to get things to work, but the `_activation_map` needs an ELU component as well.  Let me know if you want some other information.

jwsmithers: I guess the easiest thing would be to see how it's saved in the arch.json file when you call it like that. Would you happen to have one of those lying around?

demarley: I put an example file here:
`/afs/cern.ch/user/d/dmarley/public/forLWTNN/zprime_750-1-2TeV_77_31May_21Jun2017_model.json`

jwsmithers: Thanks!

Interesting, the `elu` is added sort of like another layer. It could be because it's an "advanced activation layer". I'll play around a bit and see...

demarley:
Hi @dguest and @jwsmithers is this still useful to you?  
I accidentally did all of these edits on my master branch and now I want to change some things.  I can delete this and submit a new one if helpful.

dguest:
Yes this is definitely useful! I think I got a bit distracted by this because I realized that we might need to rewrite some of the internals to properly support advanced activation functions, but for now I think we should just merge what you have.

I would really like to see an assert in the converter which warns people if they use parameters that aren't supported. You can also change the README (which currently says we don't support ELU) to say we have partial support and link to issue #77.

demarley:
Hey @jwsmithers and @dguest,
I'll leave everything as it is and you can investigate it when you have time, I just wanted to make sure it was still useful.  
I will begin implementing your suggestions Dan:

- [ ] Update README
- [ ] `assert` to warn people about `alpha` value

Issue #72, opened 2017-12-01
Maintain tensorflow and theano parameter saving between versions


jwsmithers:
From an email with Andreas Sogaard he reported an error when trying to convert a NN.

```
>>> File "/afs/cern.ch/work/a/asogaard/private/lwtnn/converters/keras_v2_layer_converters.py", line 52, in _normalization_parameters
>>>     gamma = layers['gamma'+BACKEND_SUFFIX]
>>> KeyError: 'gamma:0'

which appears to be related to the converters reading of the HDF5 weights file. The `layer_group` and `layers` in question are

>>> layer_group = <HDF5 group "/classifier/batch_normalization_1" (1 members)>
>>> layers = {'batch_normalization_1_1': array(['beta:0', 'gamma:0', 'moving_mean:0', 'moving_variance:0'], dtype='<U17’)}

I have attached the outputs from json.tool and h5ls for the architecture- and weights files, resp. They look pretty healthy to me, e.g. 

$ cat output_h5ls.txt | head -9
/                        Group
/classifier              Group
/classifier/batch_normalization_1 Group
/classifier/batch_normalization_1/classifier Group
/classifier/batch_normalization_1/classifier/batch_normalization_1_1 Group
/classifier/batch_normalization_1/classifier/batch_normalization_1_1/beta:0 Dataset {10}
/classifier/batch_normalization_1/classifier/batch_normalization_1_1/gamma:0 Dataset {10}
/classifier/batch_normalization_1/classifier/batch_normalization_1_1/moving_mean:0 Dataset {10}
/classifier/batch_normalization_1/classifier/batch_normalization_1_1/moving_variance:0 Dataset {10}

for 10 input variables, but what do I know. I am saving the model in a setup with the following package versions:
  Python version: 2.7.13 (so not python3)
  Numpy version: 1.13.1
  Keras version: 2.0.8
  TensorFlow version: 1.3.0
```

However, it actually looks like the gamma:0 does exist. This will need to be looked into. He also sent architecture and weights file which I won't attach here.

dguest:
I wonder which group it's looking for `gamma:0` in though. 

I don't remember exactly which versions of Keras and tensorflow we tested this with before: did we use Keras 2 with the tensorflow backend yet?

jwsmithers:
For the Keras 2 update this was for tensorflow and theano. I thought we had a unit test for tensorflow, but I don't remember playing with the travis CI to do this.
[...]

PR #74, opened 2017-12-12
Tweak the conversion script to be independent of naming conventions

jwsmithers:
In the past we assumed a set path for layer names. But if someone adds a "/"
then this breaks the conversion. Now we do a recursive search through each
level of the HDF5 path to see if it's a Group or Dataset. It stops and
returns the dataset.

### DIFF
--- a/converters/keras_v2_layer_converters.py
+++ b/converters/keras_v2_layer_converters.py
@@ -13,6 +13,7 @@
 import numpy as np
 import sys
 from keras_layer_converters_common import _activation_map
+import h5py
 
 def _send_recieve_meta_info(backend):
     global BACKEND_SUFFIX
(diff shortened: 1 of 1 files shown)


jwsmithers:
Solves #72 which turned out to not really be what the issue was about.

dguest:
Cool thanks for fixing this josh. It seems a bit weird that we need to do a recursive search of groups to find matches (and I worry about how this could break): shuldn't there be a unique path determined by the architecture file?

Anyway, I guess for now we focus on something that works and start thinking about how to clean up the converters in general.

jwsmithers:
I couldn't think of a foolproof way to always get the path to the dataset. As Andreas's example showed there might be some unknown string attached to the path that doesn't show up in the "name" parameter of the config.

The only immediate way I can see this breaking is if in some hdf5 config file you have a group mixed in with a dataset. I.e.:
```
batch_normalization_2
gamma
beta
```

But this would be downright weird and I don't _think_ hdf5 supports this.

The advantage of this method is that we really don't care how people name the layers, or even how tensorflow vs theano vs sklearn will save the parameters which is quite useful.

dguest:
hdf5 lets you mix groups and datasets, but I agree that I haven't seen the structure you're talking about. What I don't understand is how Keras finds the datasets when it loads them. I tried to figure this out based on the topology module: 

https://github.com/fchollet/keras/blob/master/keras/engine/topology.py

but it's not exactly clear from a quick glance

PR #75, opened 2017-12-13
Add sklearn2json converter

demarley:
Converts a simple MLPClassifier from scikit-learn to the proper JSON format.  
Some of the structure is hard-coded because I don't think the MLPClassifier has the same flexibility as Keras, but this could be wrong.

### DIFF
--- a/converters/sklearn2json.py
+++ b/converters/sklearn2json.py
@@ -1,0 +1,267 @@
+"""
+Created:      1  December 2017
+Last Updated: 12 December 2017
+
+Dan Marley
+daniel.edison.marley@cernSPAMNOT.ch
+Texas A&M University
+
+Justin Pilot
+UC Davis
+
+---
+
+Scikit-learn model -> lwtnn converter
+
+Convert simple sklearn MLP Classifer to proper format for lwtnn framework.
+Generate JSON file with variables (inputs to NN) with offset and scale values.
+  Based on converters/keras2json.py:   https://github.com/lwtnn/lwtnn
+
+
+WARNING: Some values are hard-coded due to my unfamiliarity 
+         with the sklearn MLPClassifier interface
+         http://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPClassifier.html
+
+To run and show list of options:
+   python converters/sklearn2json.py --help
+
+NB:
+   scale/offset defined from "StandardScaler"
+     http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html
+     :: offset = -mean                   (StandardScaler.mean_)
+     :: scale  = 1 / standard_deviation  (1/StandardScaler.scale_)
+"""
+import sys
+import json
+import numpy as np
+from argparse import ArgumentParser
+from sklearn import svm, metrics, preprocessing
+from sklearn.externals import joblib
+
+
+class Sklearn2json(object):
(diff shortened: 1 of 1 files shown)


dguest:
Hi @demarley, 

First off, thanks for this! We welcome any new converters even if they are a bit rough around the edges. Have you verified that this is doing what you think it should be doing (i.e. that the output is the same as you get from sklearn on a simple test pattern)? If so I'll merge it.

demarley:
Hey @dguest,
It seems the output is _slightly_ different than the predicted values from scikit-learn.  
I'm following up to make sure all of my inputs are correct and such.  
Please hold off on merging until I confirm it is working as expected.

demarley:
Validated that the output from scikit-learn and lwtnn are the same.  Please merge when you are ready.

dguest:
OK, thanks for the contribution!

I'll merge this now, but I should note that with the keras converters we always managed to get agreement between keras and lwtnn out to the precision of the cout statements. If you're seeing something with less precise agreement, it's possible that something isn't defined the same way between sklearn and keras, or that something else is wrong.

demarley:
Hey @dguest, 
From what I can tell the disagreement between python/lwtnn seems to be due to float/double precision (the ntuple values are stored as floats, and lwtnn uses doubles) and at more significant digits than anyone would "reasonably" use (...perhaps).  
I don't see any cause for concern, but maybe there's something I've missed?  
Here's an example:

`lwtnn`: 
`0.0270938 | 0.22055  | 0.748874 | 0.00301339 | 0.000468875`

`scikit-learn`:
`0.027090  | 0.220560 | 0.748860 | 0.003013    | 0.000469`

jwsmithers:
Hi Dan, 
I remember seeing something like this for the BN layer and it came down to me not implementing it exactly right. (The epsilon parameter wasn’t implemented correctly or something) Afterwards I was able to get exact matching up to like 5 or 6 decimal places.
The first number here is the problem which is a bit harder to justify with rounding/precision I think.
But I also could be wrong... :)

demarley:
Hey @jwsmithers ,
Thanks for the info!  
I will look into how the layers are implemented more closely and see if I can find some discrepancies.  
By "first number" are you referring to the `0.0270938` vs `0.027090`?  These agree to 5 decimal places, which should be okay, I hope?

dguest:
Yeah, I agree with josh here: this would register as "wrong" in most of our tests.

For comparison, we typically set `warn_threshold=0.000001` meaning yours would be firing some warnings in the CI tests (which we don't see). So what you have might be fine for most applications, but the disagreement isn't something we normally see and thus I'd be suspicious.

demarley:
Thanks @dguest for the clarification.  
I'll begin looking into issues with the converter and post a new PR when I find a fix.

dguest:
It's also possible that you're just seeing some round off error we've never seen before, so take what I say with a grain of salt. How deep is your network, and what kind of layers does it use?

demarley:
The network has 59 inputs, three hidden layers with 40 nodes (`relu` activations) and a 5-d output using `softmax`.

dguest:
Weird. That doesn't seem terribly abnormal. I wonder if you also get a discrepancy if you use e.g. a dummy input with no hidden layers, 5 inputs, and 5 softmax outputs. I guess I could believe that round off error could accumulate over that network, but if that's what's happening I'd expect better agreement from a very simple one.

demarley:
This is certainly something I will check to look for any errors!

PR #76, opened 2017-12-13
Add basic unit test to check conversion process

jwsmithers:
This uses the lstm functional unit test data.
It then creates an input file and compares to the downloaded one.
It then creates the network file using the new input file and compares
to the old one.
In each case a diff is performed on new an old. If it's empty the test
is successful.


dguest:
Hmm, so this seems like a nitpick but I'm not super happy to add a 450 KB data file to the main repository. I know it's not huge, but the entire git repository is 504 KB at this point, i.e. it's doubling the size of the repository in one commit.

dguest:
Just to elaborate a bit: would it be possible to check `basic_lstm_network_output.json` into the same repo where we put the rest of the data and then squash these commits? In general I think we should avoid using the `data/` directory.

jwsmithers:
Yup, I agree with that and new this comment was coming :). I'll change it later today.

PR #79, opened 2017-12-13
Return 1 if tests fails

jwsmithers:
As Dan pointed out the old test never actually returned 1 if there was a failure.
Whoops.
This fixes that.

### DIFF
--- a/tests/check-conversion.sh
+++ b/tests/check-conversion.sh
@@ -53,17 +53,6 @@ cd $(dirname ${BASH_SOURCE[0]})
 # ________________________________________________________________________
 # main test logic
 
-# Function to check if file is empty (as it should be) or not
-function getFileSize(){
-  if [[ -s $1 ]];
-  then
-    echo "ERROR!  Diff file is NOT empty"
-  else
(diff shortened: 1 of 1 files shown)


tests/check-conversion.sh
```diff
@@ -53,17 +53,6 @@ cd $(dirname ${BASH_SOURCE[0]})
 # ________________________________________________________________________
 # main test logic
 
-# Function to check if file is empty (as it should be) or not
-function getFileSize(){
-  if [[ -s $1 ]];
-  then
-    echo "ERROR!  Diff file is NOT empty"
-  else
```
dguest: You could just add a `return 1` here. The script is set (see the `set -eu` above) so that a function that returns 1 (an error) will cause the script to quit with an error.

jwsmithers: So interestingly enough that wasn't working. It returns 1 and exits, but for some reason decides to not print an error message (Yes, I put the return after the print ;) ). I decided to go for a slightly roundabout where we actually get see error message.

PR #84, opened 2018-08-21
Fix how CMake installs converters

dguest:
The permissions for some converter scripts were set wrong by the CMake
installer. In addition to fixing this, I removed some old converter
scripts that aren't being used (so far as anyone knows).


dguest:
@jwsmithers, just so I have another sane person's opinion, do you think this makes sense?

jwsmithers:
For what it's worth, those indeed look like sane changes to me. I don't know of or why anyone is using Julian and/or Agile converters anymore.

For the permissions, umm, sure? It looks like they're not executable anymore if I'm reading it correctly?

dguest:
@jwsmithers, `PROGRAMS` are executable, `FILES` are just readable. Thanks for taking a look.

PR #86, opened 2018-10-29
added swish activation

aghoshpub:


dguest:
Hi @aghoshpub, I think this should pretty much work as it stands, the tricky part is going to be to get the converters to recognize whatever your function is and to properly encode it in the json file that the C++ reads.

Unfortunately if it's not a standard Keras function this might be a bit ugly. What does the json file that describes this architecture look like?

src/Stack.cxx
```diff
@@ -601,6 +602,10 @@ namespace lwt {
     return out;
   }
 
+    double nn_swish( double x ){
+    return x * nn_sigmoid(x);
```
jwsmithers: Hi @aghoshpub, from this paper https://arxiv.org/pdf/1710.05941.pdf it looks like you're missing a learnable (or even constant) parameter in the sigmoid function? 
And then as @dguest mentioned in his comment, the converter will become crucial as you need to retrieve that parameter.

aghoshpub: Right I used beta=1, but indeed I should make it more general

aghoshpub:
> Hi @aghoshpub, I think this should pretty much work as it stands, the tricky part is going to be to get the converters to recognize whatever your function is and to properly encode it in the json file that the C++ reads.
> 
> Unfortunately if it's not a standard Keras function this might be a bit ugly. What does the json file that describes this architecture look like?

In fact the json file with the architecture cannot be made with the minimal implementation of Swish that I have done currently, I need to make a class. After that it should look exactly like any other activation. It will just say "swish". I'll confirm once I'm done.

src/Stack.cxx
dguest: If you're feeling motivated think it's probably worth covering the more general case, it shouldn't be super hard to do the same thing we do with the ELU just below this.

Unless there's a strong reason to stick with the notation on that paper, I would just reuse the `alpha` parameter that we already have in the configuration structure rather than adding another parameter for `beta`.

aghoshpub: Hi @jwsmithers , in fact although the paper advises a parameter I see that it is never used in implementations. Even the tensorflow implementation has no beta parameter unless I'm mistaken:
https://tensorlayer.readthedocs.io/en/stable/_modules/tensorlayer/activation.html#swish

jwsmithers: Interesting, indeed TF sets it as 1. 
I agree with @dguest. It's probably fine either way, but a general solution would be better. Keras/TF are updated constantly, and we might not notice them updating this. It might just cause us headaches down the road.

aghoshpub: I agree, having the general (non-trainable) beta parameter makes sense.

aghoshpub:
> > Hi @aghoshpub, I think this should pretty much work as it stands, the tricky part is going to be to get the converters to recognize whatever your function is and to properly encode it in the json file that the C++ reads.
> > Unfortunately if it's not a standard Keras function this might be a bit ugly. What does the json file that describes this architecture look like?
> 
> In fact the json file with the architecture cannot be made with the minimal implementation of Swish that I have done currently, I need to make a class. After that it should look exactly like any other activation. It will just say "swish". I'll confirm once I'm done.

It comes in like this:
```
, {"class_name": "Activation", "config": {"ac
tivation": "swish", "trainable": false, "name": "activation_4"}, "inbound_nodes": [[["dense_1", 0, 0, {}]]], "name": "activation_4"},
```

as opposed to:
```
, {"class_name": "LeakyReLU", "config": {"alp
ha": 0.30000001192092896, "trainable": false, "name": "leaky_re_lu_1"}, "inbound_nodes": [[["dense_1", 0, 0, {}]]], "name": "leaky_re_lu_1"},
```

src/Stack.cxx
dguest: @aghoshpub, just to clarify: we can make this general by making it *configurable*, or we can hard code it to 1, but whether or not it's trainable isn't something we have to worry about.

jwsmithers: I agree with that.
But it might amount to the same thing (trainable vs non-trainable). It depends how it enters the architecture file written by Keras. I'm guessing there will be some parameter that it spits it associated with the swish? Or if it does what TF does, it might ignore the beta altogether and then you'll _have_ to hardcode it in the converter.

aghoshpub: Currently Keras has no Swish so I have made a custom activation, and in mine, beta is set to 1.

dguest: I think hardcoding it in the converter is fine if that's what your implementation does, although better if you can get the Keras layer to emit a value for alpha. 

What I'd rather avoid is hard-coding it in the C++ code and then adding it into the json files that the C++ reads later. We need to provide backward compatibility for anything that gets stored in the json files, those can hang around and be used for multiple years.

dguest: OK, so my proposal (unless you can come up with a clever way of writing alpha from the Keras layer you implemented) is to make a `fake_alpha_activation_layer` or something similar based on [what we currently use for ELU and LeakyReLu][1]. I would just make the function take an additional parameter which gives the alpha value you want to hard code.

Again, the logic is that we can change the converters any time, but once we tell the C++ json reader to intemperate a missing `alpha` as 1, we have to support that forever.

@aghoshpub, @jwsmithers, does this sound like a good proposal?

[1]: https://github.com/lwtnn/lwtnn/blob/bbf7293332feaa84099b4827eb154d41c7c53479/converters/keras_v2_layer_converters.py#L116-L127

dguest:
OK @aghoshpub, as soon as you can verify that this is doing what you expect I can merge this and tag release 2.8. But of course we'd like to know that it's reproducing the results you're getting out of Keras first.

aghoshpub:
Hi @dguest, I made the changes like you suggested but the output of lwtnn-test-keras-functional.py is just a bunch of NANs. Did I do what you suggested?:
https://github.com/aghoshpub/lwtnn/commit/280266ad7fdabf142f762975b80cc37d3fca693e

dguest:
If the output from `lwtnn-test-keras-functional.py` is nan it's not a problem with the C++ code, that program only calls Keras.

aghoshpub:
Sorry I meant wtnn-test-lightweight-graph. I had a look at the .json file and it looks like the alpha values is not saved correctly

dguest:
hi @aghoshpub, this looks nice! 

Do you mind if I squash the this into one commit when I merge it (since it's quite a few commits)? Also please ping me when you think this is ready to merge.

aghoshpub:
Hi, Yes they should be squashed into 1 to make it easier to read the changes later. I'll go over the changes one time to make sure I didn't add redundancies. In particular, I added a file with Custom layers scripts/CustomLayers.py to help test. Also, the changes I have made are only to the functional API scripts. Do you suggest I make changes to any other script?

scripts/CustomLayers.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
```
dguest: You only need this line if you're going to make the file executable. It should definitely _not_ be executable, though: if the user puts this directory in their `$PATH` we don't want this to be an option.

converters/keras_v2_layer_converters.py
```diff
@@ -126,6 +126,23 @@ def func(h5, layer_config, n_in, layer_type):
         return pars, n_in
     return func
 
+def fake_alpha_activation_func(activation_name):
```
dguest: You might want to call this `trainable_alpha_activation_func` or something, there's no longer anything fake about the alpha.

converters/keras_v2_layer_converters.py
```diff
@@ -126,6 +126,23 @@ def func(h5, layer_config, n_in, layer_type):
...
+    def func(h5, layer_config, n_in, layer_type):
+        """Store activation 'beta' parameter of swish, referred to as 'alpha' in lwtnn"""
+        layer_group = h5[layer_config['name']]
+        layers = _get_h5_layers(layer_group)
+        alpha = layers['beta'+BACKEND_SUFFIX]
```
dguest: you _could_ also make the `'beta'` here this an argument to the `trainable_alpha_activation_function`, so the function call would be something like

```
_trainable_alpha_activation_function('swift', alpha_parameter_name='beta')
```

scripts/CustomLayers.py
aghoshpub: Right

aghoshpub:
I made the changes and pushed. I don't foresee any further changes from my side.

Issue #87, opened 2019-02-11
How does our swish function compare to keras-contrib version


dguest:
@aghoshpub added a swish function here, but apparently [it exists in keras-contrib][1]. There's probably no good reason to maintain it in both places.

[1]: https://github.com/keras-team/keras-contrib/blob/master/keras_contrib/layers/advanced_activations/swish.py

aghoshpub:
Looks like it's doing the same thing as mine.

 I had to add ```self.__name__ = 'swish'``` to get it to save the architecture on my Keras, I'm not sure how this one can get away without doing that. Perhaps a newer version of Keras doesn't require it anymore.

dguest:
OK, well my fault for not knowing about keras-contrib when we added this. 

We should try out their version at some point, I'm sure they'd be more than happy to accept a merge request if the layer that's given there doesn't save.

aghoshpub:
I tried loading a pretrained model with this architecture (with minimal changes to match the names, initialisation structure of my Swish) and it seems to work. Even without ```self.__name__ = 'swish'``` . So I think we can get rid of ours and just use the keras contrib version. I believe from lwtnn side the two will look identical, but it has to be tested. I'll do it when I find some time.

When I was making my swish function I didn't find this keras-contrib implementation, if it was already there, perhaps it was hidden away in some file with a not so obvious name. Google searches pointed only to SO back then.

dguest:
OK, well if you can confirm please make a PR to remove it. Also, should you feel so inclined, you sholud add yourself and your ORCID to [this list][1].

[1]: https://github.com/lwtnn/lwtnn/blob/master/.zenodo.json#L3-L33

aghoshpub:
So the two architectures are definitely do not look identical to lwtnn. I've been trying to find the differences we have to address. I'm not sure yet how much we need to change lwtnn to accommodate this.

dguest:
Do you think the changes could be accommodated on the converter side?

aghoshpub:
Yes I think so. The differences are only in the way the weights are saved, name of variable and the info being saved in the architecture file. All of it can be dealt with on the converter side.

aghoshpub:
One major problem with the contrib version is that if the beta is not trainable, it is not saved in the weight .h5 file. So lwtnn cannot pick it up from there. It's saved instead in the architecture .json file.

dguest:
I think we should be doing two things here:
[...]

PR #90, opened 2019-03-25
Update Swish to support Keras-contrib implementation

aghoshpub:
New swish implementation : [Keras-contrib](https://github.com/keras-team/keras-contrib/blob/master/keras_contrib/layers/advanced_activations/swish.py).

Models trained in the old swish implementation can be easily translated to the new one within keras. So LWTNN doesn't need to continue supporting the old version. The C++ side is unchanged, changes only at the converter side.

Choosing _not_ to make `Keras-contrib` a dependency, so the swish implementation is copied into `CustomLayers.py`. It will _not_ be automatically updated with an update in `Keras-contrib`.

Solves https://github.com/lwtnn/lwtnn/issues/87.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -29,6 +29,10 @@
             "name": "Daniel Edison Marley",
             "affiliation": "Texas A&M University",
             "orcid": "0000-0002-6290-078X"
+        },
+        {
+            "name": "Aishik Ghosh",
+            "orcid": "0000-0003-0819-1553"
         }
     ]
 }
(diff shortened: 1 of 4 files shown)


aghoshpub:
We could consider removing [_alpha_activation_func](https://github.com/aghoshpub/lwtnn/blob/master/converters/keras_v2_layer_converters.py#L116)
because [_trainable_alpha_activation_function](https://github.com/aghoshpub/lwtnn/blob/master/converters/keras_v2_layer_converters.py#L129)
Is a generalised version of it.

PR #92, opened 2019-05-12
Fixed compiler warnings and README typo

VukanJ:
- Marked overriding methods as 'override'
- Added more compiler warning options
- Added default virtual destructor for ISource base class (otherwise a warning is generated)
- Fixed missing initializer warning for ActivationConfig in Stack.cxx by setting alpha to zero for initialization
- ~Set CXX standard to 17, but 11 is still minimum required~
- Muted compiler warnings coming from Eigen and Boost as they cannot be fixed by code in this project (See CMakeLists.txt)
- Fixed typo in README.md

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ endif()
 message( STATUS "Using build type: ${CMAKE_BUILD_TYPE}" )
 
 # Set the warning flag(s) to use.
-set( CMAKE_CXX_FLAGS "-Wall -pedantic -Wextra -Wcomment -Wunused-value -Wsuggest-override" )
+set( CMAKE_CXX_FLAGS "-Wall -pedantic -Wextra -Wcomment -Wunused-value" )
 
 # Turn off the usage of RPATH completely:
 set( CMAKE_SKIP_RPATH ON )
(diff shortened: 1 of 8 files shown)


VukanJ:
I don't quite see why cmake throws an error if Eigen and Boost libs are marked as SYSTEM headers, so I reverted it for now (see 4761fe4). At least lwtnn warnings should be fixed by the previous two commits.

dguest:
Hi, this looks great! I was about to comment that I'd like to add some tests to the CI to build on both C++11 and C++17, but I see you switched back to C++11 now. Can you comment as to why?

VukanJ:
I changed it back because I did not know whether you have good reasons not to switch to C++17. I can set it back to C++17 if you like ;)

dguest:
I don't have a _great_ reason not to switch to C++17, but I wrote that it depends on C++11 in the README and breaking compatibility with the compiler would probably constitute a major version update. 

I think I should be able to get travis to build against both standards pretty easily though (the obvious place to start is hacking in another option in [tests/travis-build-cmake.sh](../blob/master/tests/travis-build-cmake.sh). But I can merge this as it stands since it doesn't seem to break anything.

dguest:
Thanks @VukanJ! I've merged these changes in. I'll follow up on adding C++17 to the build matrix at some point, I guess it's already 2019 so we should probably support the latest C++ standard.

Issue #102, opened 2020-06-08
Templating importent classes


benjaminhuth:
Hello,

It would be very useful if one could choose the underlying data-type which lwtnn uses. E.g., one could use single precision numbers, or even differentiable types. I tried this in [a fork of lwtnn](https://github.com/benjaminhuth/lwtnn).

As a starting point for the discussion, the following snippet shows how I approached this: 

- LWTNN right now:
   ```c++
    class Stack
    {
        // something with double...
    };
   ```
   
- Proposal:   
   ```c++
    template<typename T>
    class StackT
    {
        // something generic
    };
    
    using Stack = StackT<double>;
   ```
The typedefs are introduced to maintain compatibility with the rest of the library. 

I also think, the interface-classes can stay unchanged, we could ensure with a ``static_assert(...)`` that the type ``T`` can be assigned with double/converted to double. 

I would propose this as a change, but maybe there are also other approaches out there which we can disucss.

Best regards,
Benjamin

dguest:
Hi,

Thanks for making an issue, and sorry for the slow response. I guess the only thing I'm unsure of is the naming convention, but I don't know of any more standard ways to deal with this.

lukasheinrich:
@dguest you mean `StackT` and `Stack`?

dguest:
@lukasheinrich yeah. I guess most people would be doing this the other way around: the template would have the shorter name and the `Stack` would be `StackD` or something. But obviously we can't do that without breaking the existing code. So if no one has a better idea I'm fine with this.

dguest:
I'm also wondering if there is any reasonable way to hide the 

```C++
#include <Eigen/Dense>
```

From e.g. the `LightweightNeuralNetwork.hh` header with some kind of pImpl tricks. In the current code most of the ATLAS facing code doesn't include all the headers.

I can think of some rather convoluted solutions like having the template classes inherit from a base class, not sure if anyone has a better idea.

benjaminhuth:
Yeah the problem with the templated classes is that you always have to include the whole implementation. Maybe something like the following? (I think this is in fact pImpl).

```c++
// MyClassT.hpp
#include <Eigen/Dense>

template<class T>
class MyClassT
{
    public: 
        void do_something(T t) { /* use Eigen ... */ }
};
```
```c++
// MyClass.hpp

template<class T> class MyClassT;

class MyClass
{
    public: 
        void do_something(double t);
    private: 
[...]

PR #103, opened 2020-08-06
Templating important classes

benjaminhuth:
The discussed changes from issue #102.


dguest:
Hi @benjaminhuth, what's left to do here? Overall this looks very nice, I'm trying to decide when we should merge it.

benjaminhuth:
Hi, 
from my perspective nothing. But I thought there would be for sure some things to change in such a large update, so made a WIP PR. I think I misunderstood the concept of a WIP PR a bit...

dguest:
Hmm, github changed their interface and I didn't notice that the CI pipelines are all broken now... I should fix these before we make such a big change.

dguest:
I'm trying to figure out how to run the Travis pipelines on this merge request. It seems like something broke when microsoft started pushing actions and apps.

dguest:
Hi @benjaminhuth, can you try to rebase this branch to the current master? I just managed to get the CI pipeline running again, once you rebase it should work for you as well.

include/lwtnn/LightweightGraph.hh
```diff
@@ -42,19 +42,21 @@
  Eigen objects and Eigen objects back to std::maps. For the
  underlying implementation, see Graph.hh. */
 
-#include "lightweight_network_config.hh"
+#include <memory>
+#include <lwtnn/lightweight_network_config.hh>
```
dguest, line 46: This might be a bit of a nitpick, but it seems like changing these to `<...>` from `"..."` is adding commit noise. Is there a good reason to make this change?

include/lwtnn/Stack.hh
```diff
@@ -1,360 +1,52 @@
...
-
-  // ______________________________________________________________________
-  // Activation functions
-
-  // note that others are supported but are too simple to
-  // require a special function
+#ifndef LWTNN_STACK_HH
+#define LWTNN_STACK_HH
+
+#include <lwtnn/generic/Stack.hh>
```
dguest, line 4: Same comment here with the angle brackets include. I'd rather keep it consistently using quotes for anything in this library, angle brackets for system libraries.

include/lwtnn/generic/Graph.hh
```diff
@@ -0,0 +1,173 @@
+#ifndef LWTNN_GENERIC_GRAPH_HH
+#define LWTNN_GENERIC_GRAPH_HH
+
+#include <vector>
+#include <unordered_map>
+#include <set>
+
+#include <lwtnn/generic/eigen_typedefs.hh>
+#include <lwtnn/generic/Source.hh>
+#include <lwtnn/NNLayerConfig.hh>
```
dguest, line 10: Also here with the angle bracket includes

include/lwtnn/generic/Graph.txx
```diff
@@ -1,60 +1,71 @@
-#include "lwtnn/Graph.hh"
-#include "lwtnn/Exceptions.hh"
-#include "lwtnn/Stack.hh"
+#include <lwtnn/generic/Stack.hh>
+#include <lwtnn/generic/Graph.hh>
+#include <lwtnn/Exceptions.hh>
```
dguest, line 4: This might be the worst offender.

include/lwtnn/generic/InputPreprocessor.hh
```diff
@@ -1,50 +1,59 @@
...
 #include <map>
 #include <vector>
 
-namespace lwt {
-
-  using Eigen::VectorXd;
-  using Eigen::MatrixXd;
+#include <lwtnn/lightweight_network_config.hh>
+#include <lwtnn/Exceptions.hh>
+#include <lwtnn/generic/eigen_typedefs.hh>
```
dguest, line 9: again with the angle brackets

include/lwtnn/generic/InputPreprocessor.txx
```diff
@@ -1,14 +1,21 @@
-#include "lwtnn/InputPreprocessor.hh"
+#include <lwtnn/generic/InputPreprocessor.hh>
```
dguest, line 1: angle brackets

include/lwtnn/generic/LightweightNeuralNetwork.txx
```diff
@@ -0,0 +1,114 @@
+#include <lwtnn/generic/LightweightNeuralNetwork.hh>
+#include <lwtnn/generic/InputPreprocessor.hh>
+#include <lwtnn/generic/Stack.hh>
+#include <lwtnn/generic/eigen_typedefs.hh>
```
dguest, line 4: angle brackets include

include/lwtnn/generic/Source.hh
```diff
@@ -0,0 +1,54 @@
+#ifndef LWTNN_GENERIC_SOURCE_HH
+#define LWTNN_GENERIC_SOURCE_HH
+
+#include <lwtnn/generic/eigen_typedefs.hh>
```
dguest, line 4: another one

include/lwtnn/generic/Stack.txx
```diff
@@ -0,0 +1,795 @@
+#include <lwtnn/generic/Stack.hh>
```
dguest, line 1: use quotes here

src/LightweightGraph.cxx
```diff
@@ -1,157 +1,43 @@
-#include "lwtnn/LightweightGraph.hh"
-#include "lwtnn/InputPreprocessor.hh"
-#include "lwtnn/Graph.hh"
-#include <Eigen/Dense>
+#include <lwtnn/LightweightGraph.hh>
+#include <lwtnn/generic/LightweightGraph.hh>
+#include <lwtnn/generic/InputPreprocessor.hh>
+#include <lwtnn/generic/Graph.hh>
+#include <lwtnn/generic/eigen_typedefs.hh>
```
dguest, line 5: another example of changing the quote style. `<Eigen/Dense>` should remain as it is because it's an external library (although here I guess it's just removed).

src/LightweightNeuralNetwork.cxx
```diff
@@ -1,6 +1,7 @@
-#include "lwtnn/LightweightNeuralNetwork.hh"
-#include "lwtnn/InputPreprocessor.hh"
-#include "lwtnn/Stack.hh"
+#include <lwtnn/LightweightNeuralNetwork.hh>
+#include <lwtnn/generic/LightweightNeuralNetwork.hh>
+#include <lwtnn/generic/InputPreprocessor.hh>
+#include <lwtnn/generic/Stack.hh>
```
dguest, line 4: here too

dguest (commented):
Just to add another general nitpick: it would be nice if you used a template extension that github recognized as C++. I believe `*.tcc` should work.

Otherwise my main request is that you try to keep the use of `#include` consistent: anything referencing this library should use quotes, while external libraries should use angle brackets.

benjaminhuth:
Sorry for the delay, I was on vacation last week.

> Just to add another general nitpick: it would be nice if you used a template extension that github recognized as C++. I believe `*.tcc` should work.
> 
> Otherwise my main request is that you try to keep the use of `#include` consistent: anything referencing this library should use quotes, while external libraries should use angle brackets.

Makes sense to be consistent with these things. I changed the include notion and the file extension to `*.tcc`. Additionally to that, I added include guards to the template implementation files.

The rebase is also done, so the PR should be ready for merge, right?

dguest:
Hi @benjaminhuth, do you have any objection to squashing this MR? It seems to do exactly one thing (even if it touches a lot of code), but if you wanted to keep it as multiple commits for some reason we should discuss how that would be done.

benjaminhuth:
I didn't know yet that it is possible to squash the commits somehow... But for me its is perfectly ok to do it as one single commit!

dguest:
Alright, merged! Thanks for all the work, let's see if anything breaks!

include/lwtnn/generic/Stack.tcc
```diff
@@ -0,0 +1,800 @@
...
+
+  template<typename T>
+  T nn_tanh( T x ){
+    using std::tanh; // strange requirement from autodiff
+    return tanh(x);
+  }
+
+  template<typename T>
+  T nn_relu( T x) {
+    if (std::isnan(static_cast<double>(x))) return x;
```
dguest, line 701: Hi @benjaminhuth!

I was about to push a new release into our reconstruction code, and I got curious about this line. It _probably_ won't be a problem, but why is it needed? I assume this was put in place to work with autodiff? 

I thought the more generic thing might be to use the same trick that we use in the above functions, e.g.

```
using std::isnan;
if (isnan(x)) return x;
...
```
which should let anyone overload `isnan` however they want. Do you think that would work?

(It turns out some people are interested in using this version of lwtnn with `T = float` to try to reduce memory consumption. We also just started running a lot more NNs in our reconstruction code, so I'm trying to speed things up a bit as well.)

benjaminhuth: Hi @dguest,
Yeah that's a leftover of my autodiff-experiments. I must have missed that during my cleanup... The problem was that autodiff had no support for `std::isnan`, and this was just a quick workaround. The better way of doing this is of course the 
```c++
using std::foo;
foo(...);
```
I also used this for `std::sqrt`, etc... However, I think it would be better to do this `using` consistently for all these functions, and if I will use autodiff again someday with LWTNN, I'll just implement the correct overloads...

dguest: Hi @benjaminhuth, thanks for the quick reply! I'm about to tag a new release anyway (there were a few other things added recently that we need in ATLAS reconstruction) so I can make this change myself if you think `using std::foo; foo(...);` will accommodate autodiff.

PR #106, opened 2020-08-28
Fix an issue with shared activation functions

dguest:


dguest:
@bopingC this should fix the issues you saw.

PR #115, opened 2020-11-29
Add FastGraph

dguest:
It turns out that `lwtnn::LightweightGraph` was pretty slow in some situations because of all the `std::map<std::string, double>` lookups. This adds a new interface that avoids any strings or maps at runtime.


include/lwtnn/generic/Stack.tcc
```diff
@@ -729,11 +729,24 @@ namespace generic {
...
+    static const bool value = (
+      std::is_same<T1, T2>::value ||
+      std::is_assignable<T1, T2>::value || (
+        std::is_convertible<T2, T1>::value &&
+        std::is_floating_point<T1>::value &&
+        std::is_floating_point<T2>::value
+        )
+      );
+  };
+
```
dguest, line 745: @benjaminhuth, this is another thing that I found in the changes you made which I had to modify a bit. I found that the `std::is_same || std::is_assignable` check failed when I was using `float` as the underlying type. Since I wasn't sure exactly what case you had in mind when you added this I extended it in the minimal way possible: if both types are floating point and `T2` can be converted to `T1` the assertion passes. I don't know if `std::is_convertible` would be sufficient for the cases you tried, in which case I could probably simplify this.

benjaminhuth: Hmm, Sorry for that, I should have tested it with `float`. But obviously I thought it is sufficient as a check if it works with `autodiff`...
To explain why this `std::is_assignable` is needed: E.g. in build matrix we have this: 
```c++
T1 element = weights.at(col + row * n_inputs)
```
I think that this can either be seen as a conversion or a copy assignment, dependent on what's possible. For autodiff there is no conversion defined, just a overloaded copy assignment operator, so this would need the `std::is_assignable`. See also [here](https://godbolt.org/z/dj4zvW)

But for `float` <-> `double` we need `std::is_convertible`. I would not use `std::is_floating_point`, because that is only true for the three standard types `float`, `double` and `long double` ([cppreference](https://en.cppreference.com/w/cpp/types/is_floating_point)). So that would completely exclude user-defined types such as `autodiff::dual`.

So in the end I think this should work for all cases we need:

```c++
template<typename To, typename From>
struct conversion_check {
    static constexpr bool value = (
           std::is_same<To, From>::value
        || std::is_assignable<To, From>::value
        || std::is_convertible<From, To>::value
    );
};
```

I think it would be also nice to rename `T1` and `T2` to `To` and `From`, to make clear what we want here.

Sorry again that this wasn't as thought-out as it should have been...

PR #119, opened 2020-12-07
Add a test for the cmake installation, fix cmake bugs

dguest:
There were quite a few things missing if someone used CMake to install this library. 

This should fix a few of them. I also asked the cmake build script in the CI to install the library and then try to build another library and link against it, so hopefully we can catch more of these issues in the future.

The original CMakeLists file was written by @krasznaa, and he has far more experience with CMake than I do, so I'll wait to see if he has any suggestions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,19 +4,12 @@
 /lib
 
 # ignore common data file types
-*.yml
-*.yaml
-*.json
-*.txt
 *.h5
 *.tgz
 *.gz
 
 # except in the tests
 !/tests/data/*
-
-# include the .zenodo file
-!.zenodo.json
 
 # .pyc files
 *.pyc
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ matrix:
   include:
     # minimal build setup
     - packages:
+        - tree
         - cmake
         - libhdf5-serial-dev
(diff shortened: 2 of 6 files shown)


krasznaa:
This is one way of going about it. But if it was me, I would stop using `PUBLIC_HEADER` completely for the library.

Unfortunately that machinery only works correctly in CMake if all of the headers coming with your library are in the same directory. If you have headers put into multiple sub-directories, your best bet is to make the installation of the headers and the library independent of each other.

So I would remove all the lines having anything to do with `PUBLIC_HEADER`, and install all the headers with a single

```cmake
install( DIRECTORY "include/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}" )
```

Notice that since I wrote the configuration for lwtnn, I got into the habit of relying on CMake's own settings for the default include/library/etc. install directories.

https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

But let's not do too many things at the same time. You can just write:

```cmake
install( DIRECTORY "include/" DESTINATION "include" )
```

(Take note of how I use `/` in that code. I believe this is the correct formalism. But definitely check it yourself as well if things come out as they should...)

dguest:
Hi @krasznaa, thanks for the quick look. I wasn't sure if there was some particular reason you'd done it that way, but this is indeed much cleaner. Unless you have further comments I'll tag a release from this version.

PR #121, opened 2021-01-29
ci: Add CI with GitHub Actions

matthewfeickert:
~~This PR migrates the CI from using Travis CI to using a composition of GitHub Actions (GHA) in a GHA workflow.~~

This PR adds CI using a composition of GitHub Actions (GHA) in a GHA workflow until Travis CI is dropped.

**Suggested Squash and Merge Commit Message:**
```
* Add CI through GitHub Actions workflow
  - Run on pushes to master, pull requests, nightly CRON, and on demand
  - Test across Python 3.7, 3.8, 3.9 runtimes
  - Rename 'tests/travis-build-cmake.sh' to 'tests/build-cmake.sh'
* Add test runner Shell script
* Remove mention of Travis from repo text and add GitHub Actions badges to README
```


.travis.yml
```diff
@@ -1,68 +0,0 @@
...
-    - STANDARD=17 BUILD=./tests/travis-build-cmake.sh
-
-matrix:
-  include:
-    # minimal build setup
-    - packages:
-        - tree
-        - cmake
-        - libhdf5-serial-dev
-      env: MINIMAL=1 BUILD=./tests/travis-build-cmake.sh
```
matthewfeickert, line 27: @dguest Is this minimal test necessary? At the moment I've removed it.

.travis.yml
```diff
@@ -1,68 +0,0 @@
...
-
-notifications:
-  email: on_failure
-
-env:
-  # coverity stuff
-  global:
-   # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
-   #   via the "travis encrypt" command using the project repo's public key
-    - secure: "JpBZGZhAtJtqL8mRo4nZ7btAH+kYrLvuBZvc08FQg5IBKZFGfMfl4weeZa2Z1v/eFI1ByDuotFeBGeKA93YSaRffXyWd1pr0uhdjsU8rR/8mD9OJDAytfai8N9gVmNMlnoRg4Fwy7PVy7qB47UhdIuymZXDpcFJd0bPeYjKc5vaLXlCHrsw+px/RZpHniBl4kd8WH586To4HN4SSwTESI2/S/8W19wOAGbmlUrzmFvjdFBs8Z77CpCf3wnwcMORPFuhx06FcPtAoxAmQvo7V/3G8/Ioh8sLnmVF1/CP7Qc/2s3aIe8cLTwGSQbtVtMAZrka0gkRK6dtELMBy9bQGmY9dXhZruT50q6bT4bgk+gouPgZvMHitqWx+0T5gH+Py1VxUfV7e2B+jJA4CrlTvhADaUIJN26vRplcZnghUma1O+2SedG4aJ3zilUZqOm3K9OpfCUM4mzn1FCAcmAslO3ZqqVteK6D612WPWO7Zt6ek4hCyABtrG6YreZ2xSAVYwhma/o4aqVBrGEfecpqMGb3aAZrXeLhA+7hq4leGKvpd3OYlpjhg1KRoM26G9dr6Ac/0EIPAnJ0tf2RGey+3U9D78B9yz3ymqBZLCpjyJ1BH2RjYpHxmU0A1Ud9gxKlIJaVJ5YK9UgOg2DQZDQz4rRhpfoTVIbtTSW2+NLLr1rk="
```
matthewfeickert, line 13: @dguest There isn't a Coverity GitHub action but their website makes it seem like track can just be enabled: https://scan.coverity.com/github

matthewfeickert:

.travis.yml
dguest: Yeah that's attila's trickery to make lwtnn download and build against its own version of eigen and boost. I think it's a cool feature that can simplify deployment for some people, so I'd rather keep the tests around it.

matthewfeickert: @dguest Can I make this its own job then and have this get split off from the main test suite?

dguest: as long as it runs it's not super important where it runs. Probably worth running some run tests on this build, though, just to be sure.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+        bash tests/test-BatchNorm.sh
+        bash tests/test-dense_dropout_functional.sh
+        bash tests/test-lstm_functional.sh
+        bash tests/test-merge-graph.sh
+        bash tests/test-time-distributed-dense.sh
+        bash tests/test-gru-sequence.sh
+        bash tests/check-version-number.sh
+        bash tests/check-conversion.sh
+        bash tests/test-leaky-relu.sh
+        bash tests/test-elu.sh
```
dguest, line 60: these all start out with `#!/usr/bin/env bash`, is there some reason you have to run them with `bash`?

dguest:
I'd prefer to keep the Travis tests running until my free credit runs out. Otherwise I'm generally happy with this MR after the comments are addressed.

matthewfeickert:
> I'd prefer to keep the Travis tests running until my free credit runs out. Otherwise I'm generally happy with this MR after the comments are addressed.

Sounds good. I'll drop the commit that deletes Travis when I fix up the other parts.

.github/workflows/ci.yml
matthewfeickert: Force of habit of being explicit in scripts. I've reverted to `./` syntax.

.travis.yml
matthewfeickert: Added back to new CI

matthewfeickert:
@dguest This is ready for review again. To see an example of the CI running c.f. https://github.com/matthewfeickert/lwtnn/actions/runs/530065241

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+        libhdf5-serial-dev
+    - name: Install Python dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        python -m pip install h5py pyyaml
+        python -m pip list
+    - name: Build
+      run: ${{ matrix.build-system }}
+    - name: Run tests
+      run: |
```
dguest, line 94: Shouldn't this and the other run command just use a reference to a common list?

dguest (commented):
I only have minor comments left. Generally looks great.

.github/workflows/ci.yml
matthewfeickert: Yeah, one of the problems with GitHub Actions is that, at the moment, it isn't possible to really do much templating (one advantage that Azure pipelines really has over GHA, which is what GHA runs on in the background). Similarly, there isn't really a good way to share information between different jobs, hence I had to duplicate most of the `test` job steps to this `minimal` job. GHA is not very DRY (yet, but I think it will get there eventually).

dguest: but it is yaml right? I would have thought you could just define a yaml reference and refer to it here. Will it choke if you define a `.run-tests: &run-tests | ` type reference?

matthewfeickert: That will fail as YAML anchors aren't supported yet in GHA: https://github.community/t/support-for-yaml-anchors/16128

dguest: Holy god, they don't even conform to the yaml spec? This seems like it will end well... well ok, I guess we can copy and paste things :(

matthewfeickert: I feel similarly.

dguest: the other option is just to make one shell script call all these shell scripts. If you don't feel like doing that in this MR that's fine.

matthewfeickert: I like this idea. Done in latest push! :+1:

tests/test-runner.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -e
+set -u
+set -o pipefail
+
```
dguest, line 6: sorry I keep making suggestions, but we could also start this script out with something along the lines of `cd ${BASH_SOURCE[0]%/*}` to move into the directory where the tests are. Just a minor convenience to avoid giving a relative path for all the subsequent tests which assumes you're in the directory containing `tests/`

matthewfeickert: > relative path for all the subsequent tests which assumes you're in the directory containing `tests/`

@dguest Hm, it seems that this is actually not true as `tests/check-version-number.sh` fails if not executed from the top level of the repo given

https://github.com/lwtnn/lwtnn/blob/09d714fa1389654602052f4a69359da8faa84123/tests/check-version-number.sh#L54

c.f. this failed test build on a throw-away branch I made: https://github.com/matthewfeickert/lwtnn/runs/1818253491?check_suite_focus=true

dguest: Yeah maybe fixing that as well is really extending this merge request too far, but I suppose that's another issue that I should look into.

matthewfeickert: This is now Issue #122.

.travis.yml
dguest: ok, I guess something to look into if / when we remove `.travis.yml`.

dguest:
I'm pretty happy with this.

PR #131, opened 2021-05-03
Update URL for builtin boost

dguest:
For whatever reason the URL that we were getting boost from (in the builds that
install a builtin boost) stopped working. I updated it based on the URL I found
on 'www.boost.org', hopefully that is more robust.

Maybe @krasznaa has a reason to use the boost URL that I'm replacing here?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,9 +60,8 @@ if( BUILTIN_BOOST )
    ExternalProject_Add( Boost
       PREFIX ${CMAKE_BINARY_DIR}/externals
       INSTALL_DIR ${CMAKE_BINARY_DIR}/externals/Boost
-      URL
-      "https://dl.bintray.com/boostorg/release/1.64.0/source/boost_1_64_0.tar.gz"
-      URL_MD5 "319c6ffbbeccc366f14bb68767a6db79"
+      URL "https://boostorg.jfrog.io/artifactory/main/release/1.64.0/source/boost_1_64_0.tar.gz"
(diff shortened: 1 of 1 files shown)


dguest:
Nice that this works. I assume the old URL was simply deprecated, but since this isn't urgent we can wait a few days to see if Attila or anyone else has a comment.

matthewfeickert:
Awesome. Thanks for being super speedy with this diagnosis and fix @dguest!

krasznaa:
Unfortunately the Boost project continues to use pretty weird locations for hosting their source files. :frowning: Back when I added this code for the project, their source files were hosted on that strange place.

I was considering, even back then, to possibly make the lwtnn build depend on the source hosted by LCG instead. This particular source file you would find under:

https://lcgpackages.web.cern.ch/tarFiles/sources/boost_1_64_0.tar.gz

Note that it was exactly because of this sort of unreliability in 3rd party source file hosting that we chose to pick up source files for the ATLAS builds from CERN resources "as much as possible". In practice this means using the LCG repository and our own under http://cern.ch/atlas-software-dist-eos/externals. (We host the lwtnn sources there as well. :wink: In case GitHub would decide to change its web links at one point in the future.)

Long story short: I think this PR is good as is. But if you want to be a little safer, you may just want to use the source from the LCG webpage. Since we have an agreement with CERN IT that those sources will be available from that location for a long time. (Meaning that lwtnn tags would continue being "buildable".)

dguest:
Yeah I was wondering if we could get a more stable CERN address than this jfrog one. I've changed it to the CERN LCG url, since I have a bit more faith in that one remaining static for a while.

It's exactly the same tarball: the SHA256 is the same as the one listed on the boost website

PR #133, opened 2021-05-07
feat: Add linting with pre-commit

### LINKED ISSUE #132: Linting with pre-commit hooks?
@dguest What are your feelings about running linting with `pre-commit` hooks (both in CI and optionally with pre-commit installed locally)? I'm not suggesting trying to enforce a C++ style linter or anything like that (as much as I'd like that :wink:) but more about just using a subset of https://github.com/pre-commit/pre-commit-hooks.

If you like the idea I have a branch on my fork that is ready to go for a PR that I can demo with you as well. (c.f. https://github.com/matthewfeickert/lwtnn/runs/2522324028)

matthewfeickert:
Resolves #132 

Add common checks and linting rules with https://github.com/pre-commit/pre-commit-hooks and `pre-commit` both in terms of adding a linting GitHub Actions workflow in CI and with (optional) `pre-commit` hooks. Apply the `pre-commit` hooks globally to the project for a uniform start.

```
* Add pre-commit hooks
   - Add general checks and rules with https://github.com/pre-commit/pre-commit-hooks
* Add CI to lint with pre-commit
* Apply pre-commit rules to the codebase
```


tests/bin
```diff
@@ -1 +0,0 @@
-../bin/
```
matthewfeickert, line 1: This is apparently needed to be able to pass tests, but pre-commit fails on this if present with

```
Check for broken symlinks................................................Failed
- hook id: check-symlinks
- exit code: 1

tests/bin: Broken symlink
```

so will have to debug this later before this can go in.

dguest: yeah that's intentional: it's sort of an artifact from the `make` build method, which is in-source (unfortunately). I thought having a symlink to the output directory was better than giving all the tests a more convoluted path to follow. 

There's almost certainly a better way to point the tests to the binaries though.

matthewfeickert: > yeah that's intentional: it's sort of an artifact from the `make` build method, which is in-source (unfortunately).

Yeah, this was my assumption

> There's almost certainly a better way to point the tests to the binaries though.

I can poke around at this more next week after my vCHEP talk. :+1:

dguest: the other option is just to remove the symlink check for now and come back to it later

matthewfeickert: That's a better idea

dguest:
Because I haven't gotten around to figuring this out: if I don't actually enable these pre-commit hooks locally, what will this MR do? My impression was that it will do nothing other than make the MR CI fail if it:
 - introduce screwy whitespace
 - adds invalid json
 - does funny things with line endings

Is that correct?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.4.0
+    hooks:
+    - id: check-case-conflict
```
dguest, line 5: We don't want a repo that works on linux and breaks on MacOS? Can you enable this in Athena too?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+    - id: check-merge-conflict
+    # - id: check-symlinks
+    - id: check-json
+    - id: check-yaml
+    - id: check-toml
```
dguest, line 10: is there toml here? I didn't add it.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+    - id: check-xml
```
dguest, line 11: Can we make this fail if we have any xml at all 😁 ?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+    - id: end-of-file-fixer
+    - id: mixed-line-ending
+    - id: trailing-whitespace
```
dguest, line 14: ❤️

src/Stack.cxx
```diff
@@ -1,63 +1,63 @@
...
-      throw NNConfigurationException("sublayers in dense layer");
-    }
-  }
-
-  void throw_if_not_normalization(const LayerConfig& layer) {
-    if (layer.sublayers.size() > 0) {
-      throw NNConfigurationException("sublayers in normalization layer");
-    }
-  }
-}
```
dguest, line 63: I guess this ended up having just enough trailing whitespaces to make individual patches impossible? Cause nothing else seemed to change.

.pre-commit-config.yaml
matthewfeickert: > Check for files with names that would conflict on a case-insensitive filesystem like MacOS HFS+ or Windows FAT.

So it just safeguards across all OS. I would _love_ to have pre-commit hooks in Athena, but I have a feeling that this would get shot down faster then I could raise it.

.pre-commit-config.yaml
matthewfeickert: > Attempts to load all TOML files to verify syntax.

No, but a project that has Python in it will probably have TOML in it at some point. So this is safeguarding in the future.

matthewfeickert:
> Because I haven't gotten around to figuring this out: if I don't actually enable these pre-commit hooks locally, what will this MR do?

It will run in CI and error out if any of the pre-commit checks don't pass.


> My impression was that it will do nothing other than make the MR CI fail if it:
> 
> * introduce screwy whitespace
> * adds invalid json
> * does funny things with line endings
> 
> Is that correct?

Yes. As you had questions about each of the checks I'll just step through them and reply in the comments. I'm just summarizing what is here: https://github.com/pre-commit/pre-commit-hooks#hooks-available

.pre-commit-config.yaml
matthewfeickert: > Attempts to load all xml files to verify syntax.

Not with this pre-commit hook, but you can make a check yourself that would. ;)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+    rev: v3.4.0
+    hooks:
+    - id: check-case-conflict
+    - id: check-merge-conflict
+    # - id: check-symlinks
+    - id: check-json
+    - id: check-yaml
+    - id: check-toml
+    - id: check-xml
+    - id: end-of-file-fixer
```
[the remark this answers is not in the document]
matthewfeickert: > Makes sure files end in a newline and only a newline.

It just fails. Though if the CI is going to fail and yell at the person to change it, I'd argue that the CI might as well change it if it can. This CI doesn't, as I knew you wouldn't like that, but if you are interested in the future c.f. https://pre-commit.ci/

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+    - id: mixed-line-ending
```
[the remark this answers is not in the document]
matthewfeickert: > Replaces or checks mixed line ending.
> 
> * --fix={auto,crlf,lf,no}
>    - `auto` - Replaces automatically the most frequent line ending. This is the default argument.
>    - `crlf`, `lf` - Forces to replace line ending by respectively CRLF and LF.
>       - This option isn't compatible with git setup check-in LF check-out CRLF as git smudge this later than the hook is invoked.
>    - `no` - Checks if there is any mixed line ending without modifying any file

Yeah, we can set it to this to `lf`.

src/Stack.cxx
matthewfeickert: Nope, this was a DOS file, so I ended up running `dos2unix` on it in a61a4f6 so that `mixed-line-ending` would pass.

.pre-commit-config.yaml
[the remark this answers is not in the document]
matthewfeickert: Added in a035aa3

.pre-commit-config.yaml
dguest: I don't like it. We should think about linting toml when / if we add toml.

.pre-commit-config.yaml
dguest: I really don't like it. I can see an argument for adding toml at some point in the future but this seems extremely speculative. I'm sure there are some awesome things written in xml but I can't think of any that would belong in this project right now.

.pre-commit-config.yaml
matthewfeickert: removed in 8f3bcaab3506c8698ecc252f862624b700d9964f

.pre-commit-config.yaml
matthewfeickert: removed in ae0aba6c75da464d5bfd6ab4ccfa36fb3eb684be

dguest (commented):
Looks good.

PR #137, opened 2021-06-29
test: Add test build against ltwnn/build-base gcc v11 image

### LINKED ISSUE #136: Test build against gcc11
Some ATLAS friends [reported problems building in GCC 11](https://its.cern.ch/jira/browse/ATLINFR-4147). The error is something about us using `size_t` outside the `std::` namespace.

More generally, it might be good to have a few images around with custom compilers installed, to catch issues like this. @matthewfeickert might be able to take a look. If it's easy to add this to CI that would be awesome.

matthewfeickert:
Resolves #136 

This PR adds **testing** against `gcc` `v11` through the use of the [`lwtnn/build-base:gcc-v11.1.0` image](https://github.com/lwtnn/build-base) and is able to reproduce the build failure described in Issue #136. Fixing the build error is out of scope for this PR and should fixed in a follow up PR once this PR has been merged.

```
* Add CI stage to test the build against gcc v11 using the ltwnn/build-base images
   - c.f. https://github.com/lwtnn/build-base
* Run the test suite using the C++17 standard
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,3 +81,39 @@ jobs:
     - name: Run tests
       run: |
         bash tests/test-runner.sh
+
+  build-base:
+
+    runs-on: ubuntu-latest
+    container: lwtnn/build-base:gcc-v${{ matrix.gcc-version }}
+    strategy:
+      matrix:
+        gcc-version: [11.1.0]
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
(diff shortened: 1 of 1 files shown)


matthewfeickert:

dguest:
This is pretty awesome. As I said in #139 I probably would have fixed the issue first and then added the test to avoid any ❌s one the master branch (github will email me if master fails, not sure who else...).

I think its out of scope for this MR, but it would also be cool to add C++20 to the matrix since most of that is supported in gcc11.

PR #142, opened 2021-07-01
feat: Set range of valid CMake versions

matthewfeickert:
If only one CMake version is given to `cmake_minimum_required` then CMake will force that version to be used, even if a more updated CMake version is available. So instead of providing only a single CMake version, provide a range of supported CMake versions.

c.f. [An Introduction to Modern CMake](https://cliutils.gitlab.io/modern-cmake/chapters/basics.html) by Henry Schreiner (@henryiii)

```
* If only one CMake version is given to cmake_minimum_required then
  CMake will force that version to be used, even if a more updated CMake
  version is available. So instead of providing only a single CMake
  version, provide a range of supported CMake versions.
   - c.f. An Introduction to Modern CMake by Henry Schreiner
   - https://cliutils.gitlab.io/modern-cmake/chapters/basics.html
* List build tools available as step in CI for debugging spot checks
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,11 @@ jobs:
         libhdf5-serial-dev
         libboost-dev
         libeigen3-dev
+    - name: List build tools available
+      run: |
+        gcc --version
+        g++ --version
+        cmake --version
     - name: Install Python dependencies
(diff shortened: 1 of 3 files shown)


matthewfeickert:

PR #143, opened 2021-07-01
test: Add requirements.txt for testing Python dependencies

matthewfeickert:
This PR simply adds a `tests/requirements.txt` that contains the Python dependencies for testing with the lower bounds given by the oldest version empirically found to install from a wheel in a Python 3.6+ virtual environment. This `requirements.txt` is then used to install the test dependencies in CI and the `README` is updated to recommend dependency installation through it.

Additionally, links to the `h5py` PyPI page are added to the `README` and `requirements.txt` linting is turned on in the pre-commit hooks.

```
* Adds a tests/requirements.txt that contains the Python dependencies for testing
   - Lower bounds determined empirically by oldest release with viable wheels for modern Python 3
* Install from tests/requirements.txt in CI
* Recommend installing test dependencies from tests/requirements.txt in README
* Update pre-commit hooks and add requirements-txt-fixer linter pre-commit hook
* Link to h5py PyPI page in README
* Remove PyYAML from test requirements
```


matthewfeickert:

README.md
```diff
@@ -80,8 +81,14 @@ make -j 4
...
-If you have Python 3 and h5py installed you can run a test. Starting
-from the directory where you built the project, run
+To run the tests first install the required Python dependencies in a
+Python 3 environment
+
+```
+python -m pip install -r tests/requirements.txt
+```
+
+Starting from the directory where you built the project, run
```
dguest, line 91: I'm a bit confused because I don't think this project requires yaml. My understanding is that yaml is a requirement from the pre-commit hooks, but I guess that package should cover yaml then, no?

matthewfeickert: Let me check why this exists then. I'll go back through the commit history to see where it was introduced.

matthewfeickert: @dguest `PyYAML` usage was introduced by you in 2016 in 2d295406 for reasons I'm not sure of. I'll try removing it and see if the CI passes.

dguest: I thought we were using yaml as part of the CI, which is fine (we use a few things there that aren't strict requirements). I'm just uneasy making this part of the "quick start" instructions because one of the aims of this project is to have as few dependencies as possible.

matthewfeickert: No, I don't think we ever need Python to parse YAML anywhere, so I think the CI will pass without it. I'd still highly recommend keeping a `requirements.txt` though as it is important to know what your lower bounds are on versions that you support. Keeping those requirements in a `requirements.txt` is great for applications or testing purposes like this as then you have a single source of truth which everything uses.

dguest: ahhhhhh, it was part of agilepack :) 

We don't support agilpack anymore. I think it's been at least 5 years since that repository had a big banner saying it wasn't supported.

PR #144, opened 2021-07-07
Go back to truncating Sigmoid at [-30, 30]

dguest:
In e3622ddf5d8e6558 I replaced the old std::function based sigmoid function
with one that uses Eigen arrays. In doing this I dropped the checks that would
return 0 for inputs below -30 and 1 for inputs above 30. These checks weren't
strictly needed: numbers outside this range would cause under/overflows but the
result should still have been correct.

For better or worse, though, ATLAS checks for FPEs. Dropping the checks above
resulted in the FPEs reported in
https://its.cern.ch/jira/browse/ATLASRECTS-6461. This commit returns to the old
behavior which returns 0 or 1 for values outside [-30, 30].

### DIFF
--- a/include/lwtnn/generic/Stack.tcc
+++ b/include/lwtnn/generic/Stack.tcc
@@ -214,7 +214,17 @@ namespace generic {
 
   template<typename T>
   VectorX<T> SigmoidLayer<T>::compute(const VectorX<T>& in) const {
-    return 1.0 / (1.0 + exp(-in.array()));
+    using std::exp;  // strange requirement from autodiff
+    return in.unaryExpr([](T x) -> T {
+      // These checks are to prevent a floating point overflow or
+      // underflow in the exponentiation. Neither under or overflows
+      // are a problem for the output (an overflow in the exp should
+      // still return 0, and an underflow should return 1), but ATLAS
+      // runs FPE auditors and the overflows make them sad.
+      if (x < -30.0) return 0.0;
+      if (x >  30.0) return 1.0;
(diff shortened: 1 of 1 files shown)


matthewfeickert:
@dguest is is possible to add a test case for what was causing the failure earlier?

dguest:
Adding a test would probably be possible... We'd have to figure out how to turn on FPE auditing though. Without FPE auditing (which I don't think is usually enabled by default) this won't be a noticeable problem since an underflow will produce `1 / (1 + 0) = 1` and an overflow will produce `1 / (1 + inf) = 0`, i.e. exactly the desired outcome.

AnChristos:
A simple way , minimal thing to see if any FPE has occured somewhere/anywhere  for simple testing 
could look  like

https://godbolt.org/z/W5qx4bv8n
See also [1] 

Of course one could install handlers (see "GNU libc function feenableexcept() enables trapping of the floating-point exceptions, which generates the signal SIGFPE. " [1] ) 
And do it more like we do in ATLAS.


[1] https://en.cppreference.com/w/cpp/numeric/fenv

dguest:
Thanks @AnChristos, this is really helpful! I think figuring out how we should test for FPEs is probably a more complicated discussion: it will depend a lot on what input domain we want to cover, obviously we can force a lot more FPEs out of this code by feeding in extreme values. 

But since this is a bit of a regression from the FPE point of view I'm going to merge this for now. Maybe we should discuss a more comprehensive test in another issue.

PR #149, opened 2021-08-11
Migrate json parser to nlohmann_json

fwinkl:
Migrate the json parser from `boost` to `nlohmann_json`, which provides
a better JSON API, is signifcantly faster and results in much smaller
binaries (up to 30x smaller).

In particular, for an ATLAS electron DNN:
- 4x faster reading of the JSON
- `malloc`s reduced from ~5M to 2k
- 10 times smaller resident memory size

If this is a welcome change, this PR needs a bit more work, i.e. to validate the changes on all possible network configurations. I would need some advice on how to do that (the unit test coverage seems not very good so far, `test-GRU.sh` passes but it does not seem to test all code paths).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 
 # Set range of valid CMake versions to build the project.
-cmake_minimum_required(VERSION 3.1...3.20)
+cmake_minimum_required(VERSION 3.11...3.20)
 
 if(${CMAKE_VERSION} VERSION_LESS 3.12)
     cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 
 # Set range of valid CMake versions to build the project.
-cmake_minimum_required(VERSION 3.1...3.20)
+cmake_minimum_required(VERSION 3.11...3.20)
```
fwinkl, line 3: I bumped the minimum to `3.11` in order to use `FetchContent` below. But that can certainly be changed if you want to keep the minimum at `3.1`.

include/lwtnn/NNLayerConfig.hh
```diff
@@ -63,7 +63,7 @@ namespace lwt {
   struct NodeConfig
   {
     enum class Type {
-      INPUT, INPUT_SEQUENCE, FEED_FORWARD, CONCATENATE, SEQUENCE,
+      NONE, INPUT, INPUT_SEQUENCE, FEED_FORWARD, CONCATENATE, SEQUENCE,
```
fwinkl, line 66: Added `NONE` here as the `nlohmann` parser falls back to the first `enum` element in case of an unknown string. All other enums already had `NONE` as their first element, so this should hopefully be a safe change (I assume nobody uses the bare `int` values anywhere).

src/parse_json.cxx
```diff
@@ -1,344 +1,262 @@
...
-        if (v.second.count(dkey)) {
-          std::string key = v.second.get<std::string>("name");
-          defaults.emplace(key, v.second.get<double>(dkey));
-        }
-      }
+  void add_embedding_info(LayerConfig& layer, const nlohmann::json& j) {
+    for (const auto& sub: j.at("sublayers")) {
+      lwt::EmbeddingConfig emb;
+      sub.at("weights").get_to(emb.weights);
+      //emb.index = sub.second.get<int>("index"); FIXME (need an example NN file)
```
fwinkl, line 257: Was not able to test this part of the code yet. Need a NN example.

src/parse_json.cxx
```diff
@@ -1,344 +1,262 @@
...
+    switch(layer.architecture) {
+    case Architecture::DENSE:
+      add_dense_info(layer, j);
+      break;
+    case Architecture::NORMALIZATION:
+      add_dense_info(layer, j); // re-use dense layer
+      break;
+    case Architecture::MAXOUT:
+      add_maxout_info(layer, j);
+      layer.activation.function = Activation::NONE; // FIXME: to make throw_if_not_maxout() happy
```
fwinkl, line 127: The previous code set `function` to `NONE` but the NN file I tested with actually contains "linear":
```
    {
      "activation": "linear",
      "architecture": "maxout",
      "sublayers": [
```
Could we just leave the value on `LINEAR` and get rid of [this check](https://github.com/lwtnn/lwtnn/blob/84dcb5e646b631d29ef6830975a3dd7932e2a3be/src/Stack.cxx)?

dguest:
I really like this change (was never a huge fan of `boost::property_tree`), but I'm thinking it would be something like a major version bump given how much we've emphasized minimal dependencies and that this is swapping one of only two dependencies.

I'm also mostly offline until September, is this something that someone in ATLAS wants urgently?

fwinkl:
Yes, this can (and probably should) wait for a major version bump. In the meantime I can work on some of the remaining issues. But some instructions on how to more extensively test this would be needed at some point.

dguest:
I had to approve the pipelines, but pipelines/ci.yml hasn't been modified to install nlohmann, which seems to be making them all fail.

fwinkl:
> I had to approve the pipelines, but `workflows/ci.yml` hasn't been modified to install nlohmann, which means all the ci jobs can't find the json parser. We should be able to add `nlohmann` where `libboost-dev` is now (and remove boost, since we only used it for `ptree`).

I tried that but the `nlohmann-json-dev` on `bionic` is hopelessly outdated. Any objections to just get nlohmann from github also in the CI?

dguest:
Hi @fwinkl, pulling from github sounds fine to me. 

I'm a bit worried about relying on any bleeding-edge dependencies, but it's reasonable to think nlohmann has come a long way since 2018. Since we're going to bump the major release anyway I'm just wondering if we should also clean a few other things up...

nsmith-:
Have you considered [rapidjson](https://rapidjson.org/) or [simdjson](https://simdjson.org/)? For [correctionlib](https://cms-nanoaod.github.io/correctionlib/) we've been quite happy with rapidjson performance and low maintenance burden.

dguest:
Have you looked at [boot.JSON](https://www.boost.org/doc/libs/latest/libs/json/doc/html/index.html#overview)? The advantage of switching to that would be that we're not technically adding a new dependency, just requiring a new version. Of course boost is a special dependency in that sense, it's quite a monster.

PR #150, opened 2021-09-14
Add SimpleRNN layer

laurilaatu:
This pull request adds `SimpleRNN` layer to `lwtnn` following the style of other recurrent layers.

Please let me know if there are any needed changes.

### DIFF
--- a/converters/keras_v2_layer_converters.py
+++ b/converters/keras_v2_layer_converters.py
@@ -80,12 +80,15 @@ def _rnn_parameters(h5, layer_config, n_in, layer_type):
     elif "gru" in layer_type:
         elements = "zrh"
         rnn_architecure = "gru"
+    elif "simplernn" in layer_type:
+        elements = "h"
+        rnn_architecure = "simplernn"
     else:
         sys.exit("We don't recognize the layer {}"
-            .format(layer_type))
+                 .format(layer_type))
 
     layers = _get_h5_layers(layer_group)
-    n_out = layers['recurrent_kernel'+ BACKEND_SUFFIX].shape[0]
+    n_out = layers['recurrent_kernel' + BACKEND_SUFFIX].shape[0]
     submap = {}
 
     for n_gate, gate in enumerate(elements):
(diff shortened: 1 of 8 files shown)


converters/keras_v2_layer_converters.py
```diff
@@ -98,9 +101,12 @@ def _rnn_parameters(h5, layer_config, n_in, layer_type):
...
 
-    return {'components': submap, 'architecture': rnn_architecure,
-            'activation': activation_map[layer_config['activation']],
-            'inner_activation': activation_map[layer_config['recurrent_activation']]}, n_out
+    rnn_parameters = {'components': submap, 'architecture': rnn_architecure,
+                        'activation': activation_map[layer_config['activation']],
+                        'inner_activation': activation_map[layer_config['recurrent_activation']]} if rnn_architecure != "simplernn" else {'components': submap, 'architecture': rnn_architecure,
+                        'activation': activation_map[layer_config['activation']]}
+
+    return rnn_parameters, n_out
```
dguest, line 109: This might be more readable as
```
rnn_parameters = {...}
if rnn_architecture != 'simplernn':
    rnn_parameters['inner_activation'] = activation_map[layer_config['recurrent_activation']]
```
as it stands it's hard to see what the if statement is doing

dguest: also feel free to fix the spelling mistake in `architecture`...

dguest (changes requested):
Overall looks good. I had one minor comment on readability in an `if ... else` block.

Two more general comments:
- I usually try to break at 80 characters. Not all the code currently follows this and it's not essential that contributions do, but if it's easy to make the code follow this guideline I'd encourage it.
- There aren't any unit tests for this layer. I like to have unit tests that run the (keras output) -> (json) -> (apply network on example data) chain so that these things don't break in the future. Unfortunately adding a test isn't as simple as it could be, but if you raise an issue and attach:
   - The variable specification json file
   - The keras weights in h5 format
   - The network architecture json file
 
  we could use that as a starting point to adding a test.

converters/keras_v2_layer_converters.py
laurilaatu: Thanks for the comment. I fixed both. Let me know if the solution is good.

dguest:
I created #151 to keep track of the unit test we'd like to have

Issue #151, opened 2021-09-15
Add unit test for SimpleRNN


dguest:
With #150 @laurilaatu added a `SimpleRNN` layer. It would be good to have a test for this. We need:

- [ ] An example network architecture file
- [ ] The weights HDF5 file for that network
- [ ] The variable specification file for the inputs and outputs

PR #152, opened 2021-09-17
Add tests for SimpleRNN

laurilaatu:
This pull request adds tests mentioned in issue #151 based on the existing `test-lstm_functional.sh` file.

For this I needed to fix a few bugs in `tests/reg-test.py` and it seems that the output format has changed too with the values being wrapped in brackets.

Please let me know if this approach is good and if any changes are needed.


tests/reg-test.py
```diff
@@ -26,18 +26,19 @@ def _get_args():
         formatter_class=argparse.RawDescriptionHelpFormatter)
     parser.add_argument('saved_variables', nargs='?')
     parser.add_argument('-t', '--tolerance', type=float, default=0.00001)
-    parser.add_argument('-g', '--graph', action='store_true',
+    parser.add_argument('-g', '--graph', action='store_false',
```
dguest, line 29: Inverting this option seems like it should break a lot of unit tests, and indeed some of them aren't passing.

dguest (commented):
I'm sort of surprised that all these changes are needed: the structure should be identical to the lstm test. I can take a look in a few days but are you sure so much has to be changed in the `reg-test.py` script?

tests/reg-test.py
laurilaatu: Alright. I will flip it back to default true. This flag affected the used `parse` function and one of them did not work on many output formats.

laurilaatu:
Thanks for review.

After further testing in commit `5a31f51685e5d8b062d63c00650a5b8f614006ad` I noticed that the issues are caused by the `_get_dict` parser function:
`     parse = _get_output_node_dicts if args.graph else _get_dict`

With this line and the `lstm_functional.json` test output is following when piping the output according to the advice given in the test script
```
# Target file, can be generated by piping the NN output through the
# `reg-test.py` script. We also use the `reg-test.py` routine to make
# sure outputs haven't changed. See `reg-test.py -h` for more info.
```
Output:

```
./bin/lwtnn-test-lightweight-graph repo_lstm_test.json | ../tests/reg-test.py 
Traceback (most recent call last):
  File "../tests/reg-test.py", line 140, in <module>
    _run()
  File "../tests/reg-test.py", line 41, in _run
    in_dict = parse(sys.stdin)
  File "../tests/reg-test.py", line 101, in _get_dict
    key, val = line.split()
ValueError: not enough values to unpack (expected 2, got 1)

```

By using `_get_output_node_dicts` for regardless of `--graph` flag, the program runs as expected:
```
{
  "dense_1_0:": {
    "out_0": [
      0.260607
    ],
    "out_1": [
      0.213633
    ],
    "out_2": [
      0.244948
    ],
    "out_3": [
      0.280811
    ]
  }
}

```
And `./bin/lwtnn-test-lightweight-graph repo_lstm_test.json | ../tests/reg-test.py --graph ../tests/data/lstm_functional-out.json` returns `all outputs within thresholds!` same for other `merge_graph` and `highway`.

Also in commit `5a31f51685e5d8b062d63c00650a5b8f614006ad` the highway test failure happens when passing the `--graph` argument:

```
./bin/lwtnn-test-arbitrary-net repo_high_test.json | ../tests/reg-test.py --graph ../tests/data/highway-out.json 
Traceback (most recent call last):
  File "../tests/reg-test.py", line 140, in <module>
    _run()
  File "../tests/reg-test.py", line 41, in _run
    in_dict = parse(sys.stdin)
  File "../tests/reg-test.py", line 90, in _get_output_node_dicts
    odict[node_key][key] = [float(val) for val in vals]
KeyError: None

```


In addition to changes related to the parsing this PR modifies `_compare_equal` which now allows nested structures as inputs.


The tests seem to run cleanly through now after the `SimpleRNN` test files were merged and tagged.

Please let me know if you want something changed back to how it was.

dguest:
ok, I tested this out on my side. I converted your test network to something called `lwtnn.json` and ran

```
> ~1/../build/bin/lwtnn-test-lightweight-graph lwtnn.json | ~1/../lwtnn/tests/reg-test.py --graph | tee out.json
{
  "dense_1_0:": {
    "out_0": [
      0.0627666
    ],
    "out_1": [
      1.7351e-06
    ],
    "out_2": [
      0.497919
    ],
    "out_3": [
      0.439312
    ]
  }
}
> ~1/../build/bin/lwtnn-test-lightweight-graph lwtnn.json | ~1/../lwtnn/tests/reg-test.py --graph out.json
all outputs within thresholds!
```

so it seems to work on my side. I wonder if you were just missing a `--graph` argument to the `reg-test.py` function (if that's it I'm sorry the instructions were wrong...)

tests/data/simplernn_functional-out.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dense_1_0:": {
+    "out_0": 0.0627666,
+    "out_1": 1.7351e-06,
+    "out_2": 0.497919,
+    "out_3": 0.439312
+  }
```
dguest, line 7: ok, I suspect this might be the problem: when I ran `reg-test.py --graph` I got a different output, specifically with the numbers here enclosed in `[]`.

laurilaatu: The format of this file is the same as in `tests/data/lstm_functional-out.json`

laurilaatu:
> ok, I tested this out on my side. I converted your test network to something called `lwtnn.json` and ran
> 
> ```
> > ~1/../build/bin/lwtnn-test-lightweight-graph lwtnn.json | ~1/../lwtnn/tests/reg-test.py --graph | tee out.json
> {
>   "dense_1_0:": {
>     "out_0": [
>       0.0627666
>     ],
>     "out_1": [
>       1.7351e-06
>     ],
>     "out_2": [
>       0.497919
>     ],
>     "out_3": [
>       0.439312
>     ]
>   }
> }
> > ~1/../build/bin/lwtnn-test-lightweight-graph lwtnn.json | ~1/../lwtnn/tests/reg-test.py --graph out.json
> all outputs within thresholds!
> ```
> 
> so it seems to work on my side. I wonder if you were just missing a `--graph` argument to the `reg-test.py` function (if that's it I'm sorry the instructions were wrong...)

Thanks for the reply.

How does one create test outputs using `reg-test.py` like the ones found in `tests/data` if the `--graph` argument is a required one?

tests/data/simplernn_functional-out.json
dguest: hm, yeah that's weird and something I should look into...

dguest:
> How does one create test outputs using `reg-test.py` like the ones found in `tests/data` if the `--graph` argument is a required one?

The `--graph` is only for functional models, if you were doing a sequential one it would have the other format. I think it's just configured to be able to read both.

I should say that normally I make the reference dataset in `test/data` from the keras outputs, since those are the ones we're actually targeting. But I already checked that the keras and lwtnn outputs seem to agree in this case, so either one is fine.

matthewfeickert:
@laurilaatu this is a bit of an aside, but for future PRs (thanks for making these nice contributions btw!) can you please make them from **new** branches on your fork? Working on `master` on your fork is a bit of an anti-pattern as it makes it literally impossible to rebase anything without closing the PR and more difficult to update and resolve merge conflicts. 

An "easy" way to avoid having to deal with this in general is to just delete the `master` branch on your fork so that you're force to use a development branch that is different from the project default branch.

tests/data/simplernn_functional-out.json
dguest: My guess is that I made these files "by hand" from the keras outputs, since as I commented elsewhere the real target is whatever comes out of keras. I think either format will work for these files when you're comparing to the lwtnn output, the problem is just that the `reg-test.py` script creates the other format, right?

dguest:
Just to be clear, I think this is good to merge if we revert the changes to `reg-test.py`. Feel free to open some issues if you think the instructions to add tests were unclear, and my apologies if this was harder than it has to be.

tests/data/simplernn_functional-out.json
laurilaatu: Yeah, it's the output that `reg-test.py` gave. Both formats work when testing.

laurilaatu (commented):
Thanks for the review. I reverted the changes in `reg-test.py`, but left in one bugfix to print the usage. I will create another branch in my fork for future PRs.

dguest:
Thanks for your patience, merged!

PR #158, opened 2021-11-25
Improving kerasfunc2json.py compatibility

sfranchel:
Making `converters/kerasfunc2json.py` compatible with dropout layers in  `timedistributed` type of layers.

### DIFF
--- a/converters/kerasfunc2json.py
+++ b/converters/kerasfunc2json.py
@@ -21,7 +21,7 @@ import sys, os
 from keras_layer_converters_common import skip_layers
 
 def _run():
-    """Top level routine""" 
+    """Top level routine"""
     args = _get_args()
     with open(args.arch_file, 'r') as arch_file:
         arch = json.load(arch_file)
@@ -228,9 +228,7 @@ def _build_node_dict(network):
     # Remove the nodes and sources that are of type skip_layers
     removed_nodes = set()
     for node_index, node in nodes.items():
-        if node.layer_type in skip_layers:
-            removed_nodes.add(node_index)
-        elif "timedistributed" in node.layer_type and node.keras_layer['config']['layer']['class_name'].lower() in skip_layers:
+        if _is_a_skipped_node(node):
             removed_nodes.add(node_index)
         else:
             new_sources = []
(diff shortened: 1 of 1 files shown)


converters/kerasfunc2json.py
```diff
@@ -21,7 +21,7 @@
 from keras_layer_converters_common import skip_layers
 
 def _run():
-    """Top level routine"""
+    """Top level routine"""
```
dguest, line 24: We don't want to add trailing whitespace, I think @matthewfeickert's linting magic will pick up on this though

converters/kerasfunc2json.py
```diff
@@ -220,16 +220,18 @@ def _build_node_dict(network):
...
     # now we collapse the node references
     for node in nodes.values():
-            source_nodes = []
-            for source in node.sources:
-                source_nodes.append(nodes[source])
-            node.sources = source_nodes
+        source_nodes = []
+        for source in node.sources:
+            source_nodes.append(nodes[source])
+        node.sources = source_nodes
```
dguest, line 226: your indentation looks right, I'm just surprised the linter allowed what we had before

converters/kerasfunc2json.py
```diff
@@ -220,16 +220,18 @@ def _build_node_dict(network):
...
 
     # Remove the nodes and sources that are of type skip_layers
     removed_nodes = set()
     for node_index, node in nodes.items():
         if node.layer_type in skip_layers:
             removed_nodes.add(node_index)
+        elif "timedistributed" in node.layer_type and node.keras_layer['config']['layer']['class_name'].lower() in skip_layers:
+            removed_nodes.add(node_index)
```
dguest, line 234: I think here, and in the line below, we want some function like `_is_a_skipped_node(node): ... ` that executes the logic in this strong

converters/kerasfunc2json.py
```diff
@@ -242,7 +244,12 @@ def _build_node_dict(network):
...
         Apply this recursively"""
-    if node_source.layer_type not in skip_layers:
+
+    if "timedistributed" in node_source.layer_type and node_source.keras_layer['config']['layer']['class_name'].lower() in skip_layers:
+        assert len(node_source.sources) == 1
+        #@Todo: Check that this will work with two skip_layers in a row
+        return _get_valid_sources(node_source.sources[0])
+    elif node_source.layer_type not in skip_layers:
         return node_source
     else:
```
dguest, line 254: Related comment to the one above here: we really want something like

```python
if not _is_a_skipped_node(node):
    return node_source
else:
    ....
```
to avoid repeating the same logic (in slightly different ways) in two places

dguest (changes requested):
Looks pretty reasonable overall. We could consolidate your logic a bit. Let's see what the CI tells us.

dguest:
Yeah the linter picked up the trailing whitespace, as I expected. Good news is that everything else seemed to work fine.

converters/kerasfunc2json.py
sfranchel: ops, that one was a mistake...

converters/kerasfunc2json.py
sfranchel: Ok, I can put all that in a single function

dguest:
Hi @sfranchel I think we're still failing linting tests because of trailing white space. You can usually set your text editor to highlight this.

converters/kerasfunc2json.py
matthewfeickert: I didn't setup enforcement of tabs vs. spaces in the linter in PR #133. Though we can. :)

matthewfeickert:
> I think we're still failing linting tests because of trailing white space. You can usually set your text editor to highlight this.

@sfranchel to make this easy on you, my recommendation is (assuming you already have a Python virtual environment setup for running the tests) in your virtual environment just install `pre-commit` and then run it from the top level of the repo to pick up the pre-commit config:

```console
$ python -m pip install pre-commit
$ pre-commit run --all-files
```

converters/kerasfunc2json.py
dguest: Oh god, so we might have tabs all over the place... I don't really want to make yet another commit that touches hundreds of lines just for this. Is there some way we can make the linter only complain about changed lines that contain tabs?

dguest:
Merged, thanks @sfranchel!

converters/kerasfunc2json.py
matthewfeickert: For posterity, PR #159 showed us that we only had to touch 1 line. Small miracles happen. :)

PR #163, opened 2022-04-18
test:  Remove lwtnn version in tests find_package to ensure latest

matthewfeickert:
~~This PR updates the test version installed to `v2.13` to ensure the latest version is being tested.~~

This PR removes the specification of the test version, which should safely pick up the latest version for testing.

### DIFF
--- a/tests/install/CMakeLists.txt
+++ b/tests/install/CMakeLists.txt
@@ -4,6 +4,6 @@ project(TestLwtnn)
 set( CMAKE_CXX_STANDARD 11 CACHE STRING
   "C++ standard to use for the build" )
 
-find_package( lwtnn 2.12.1 REQUIRED )
+find_package( lwtnn REQUIRED )
 add_executable(test-lwtnn test-lwtnn.cxx )
 target_link_libraries( test-lwtnn lwtnn::lwtnn )


matthewfeickert:
@dguest this is ready for review.

dguest:
I guess CMake follows semantic versioning well enough that the old string still worked.

I'm sort of wondering if we should just drop [the optional version number on `find_package`][1] in this test, rather than having to increment this in one more place. If the the installation test is able to find another version something is broken anyway.

[1]: https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature

dguest:
I guess CMake follows semantic versioning well enough that the old string still worked.

I'm sort of wondering if we should just drop [the optional version number on `find_package`][1] in this test, rather than having to increment this in one more place. If the the installation test is able to find another version something is broken anyway.

[1]: https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature

matthewfeickert:
> I'm sort of wondering if we should just drop [the optional version number on `find_package`](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature) in this test, rather than having to increment this in one more place. If the the installation test is able to find another version something is broken anyway.

@dguest Yeah you make a good point, given the docs

> If no `[version]` and/or component list is given to a recursive invocation inside a find-module, the corresponding arguments are forwarded automatically from the outer call (including the `EXACT` flag for `[version]`). 

As this is also in the test suite then it should be testing what it can find anyway.

Let me change this to a removal and rebase it.

Issue #170, opened 2022-10-28
Support Keras add layers


tprocter46:
Add support for the keras "Add" layer (https://keras.io/api/layers/merging_layers/add/)

This came up while trying to convert an ONNX  model to lwtnn for preservation in Rivet, and after a brief discussion with @dguest and @matthewfeickert .

I've implemented this locally on my machine (including amending the converter script kerasfunc2json.py). Given the layer has many similarities to concatenate, I basically used that as the guide for what to do. I'm happy to start a pull request or share a `git diff` or similar to share the changes - I would have just attached the main files here but github doesn't support `.hh`, `.tcc` files etc.

I haven't yet tried to write a unit test, but I'm happy to have a go at this too, if you'd like me to.

dguest:
Yeah by all means just make a pull request.

If you want to add a unit test that would also be awesome. You can probably base it on one of the more recently edited tests in the `tests/` directory. We have a pretty unsophisticated way to deal with the binaries that the tests need as inputs: you also make a MR to the lwtnn-test-data repo and add the weights and architecture there, and then use `wget` to download them in the unit test.

PR #171, opened 2022-10-31
Keras add layer

tprocter46:
Following from Issue #170 , an implementation of the keras "add" layer - just adds (two or more) same-sized outputs together.

Also includes a test - so there's an imminent PR to the test-data repo incoming.
At the moment, the `$INPUT` variable in the test points to the branch of my fork of the test-data repo to make it work, but I can change this depending on which way round it's normal to submit PRs.


tests/test-Add.sh
```diff
@@ -0,0 +1,93 @@
...
+#  - Read in the saved file, with C++, test on a few patterns
+#  - Compare the outputs to somehting previously saved
+
+# _______________________________________________________________________
+# define inputs, outputs, and code to run
+#
+# If to add another test you'll probably have to edit this
+
+# Trained network to convert and test
+INPUT=https://github.com/tprocter46/lwtnn-test-data/raw/add-test/add_dense_functional.tgz
```
dguest, line 16: I merged your pull request in the other repository and tagged it as v10, so this can point there now.

dguest (commented):
Overall looks pretty great, thanks. A few minor comments.

tests/test-Add.sh
tprocter46: Thanks! Updated.

include/lwtnn/generic/Graph.tcc
```diff
@@ -152,6 +152,41 @@ namespace generic {
...
 
+  template<typename T>
+  AddNode<T>::AddNode(const std::vector<const INode<T>*>& sources):
+    m_sources(sources)
+  {
+    assert(sources.size() > 0);
+    m_n_outputs = sources[0]->n_outputs();
+    //Check to make sure each input layer has the same size
+    for (const auto source: sources) {
+      assert(source->n_outputs() == m_n_outputs);
```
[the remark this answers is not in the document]
tprocter46: I was trying to copy the style from the ConcatenateNode compute method, but that explanation makes sense. Thanks - useful to know!
I missed the first one originally, so it's across two commits, sorry - but fixed.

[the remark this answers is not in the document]
dguest: Yeah we're using `assert` in the `compute` methods a bit, where it makes sense to strip them out.

include/lwtnn/generic/Graph.tcc
```diff
@@ -469,7 +508,15 @@ namespace generic {
...
       m_nodes[iii] = new SumNode<T>(m_seq_nodes.at(node.sources.at(0)));
-    } else {
+    } else if (node.type == NodeConfig::Type::ADD) {
+      // build add layer
+      std::vector<const INode<T>*> in_nodes;
+      for (std::size_t source_node: node.sources) {
+        in_nodes.push_back(m_nodes.at(source_node));
+      }
+      m_nodes[iii] = new AddNode<T>(in_nodes);
+    }
```
dguest, line 518: The linter seems quite upset about this line: it has a space after the `}` and we don't allow trailing whitespace.

dguest (commented):
Everything seems alright, but you've gone and made the linter very angry, better fix that.

include/lwtnn/generic/Graph.tcc
```diff
@@ -152,6 +152,45 @@ namespace generic {
...
+    for (const auto source: sources) {
+      if(source->n_outputs() != m_n_outputs){
+        throw NNConfigurationException("All sources of an add layer must have same dimension");
+      }
+    }
+  }
+
+  template<typename T>
+  VectorX<T> AddNode<T>::compute(const ISource<T>& source) const {
+    VectorX<T> output(m_n_outputs);
```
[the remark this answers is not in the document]
tprocter46: `I haven't done that much work directly interfacing with Eigen, so I didn't know about the `VectorX::Zero` - thanks - it's definitely cleaner than before, and halves the length of the function.

I also used the opportunity to move the input.rows() call inside the assert - it probably won't make much difference, but if the assert gets stripped out then there's no point assigning an unused `size_t`, right? (though I guess the compiler might  realise this and won't actually call it anyway).

include/lwtnn/generic/Graph.tcc
tprocter46: Fixed

Issue #175, opened 2024-04-12
Support for Keras version 3.x.x


AhmedMarkhoos:
At the moment, [keras2json.py](https://github.com/lwtnn/lwtnn/blob/master/converters/keras2json.py) does not support conversion of json files from keras versions >3.x.x.

dguest:
We should definitely add it. If you have an example of a trained network you'd like to convert I might have some time to take a look at some point, or maybe someone else will.

callum-mccracken:
Adding an example keras 3.3.3 model, would also appreciate an update :)

[model.zip](https://github.com/user-attachments/files/17554328/model.zip)

callum-mccracken:
Actually scratch that last upload, that was after a little editing to try to make lwtnn work with it. Fresh version is uploaded here.

[model_run3_WH_Zdep.zip](https://github.com/user-attachments/files/17554611/model_run3_WH_Zdep.zip)

What I had changed: the keras version and backend are no longer saved in the architecture file, tried manually adding them (didn't work)

PR #177, opened 2024-10-30
Fixes for version v3.3

dguest:
I don't expect this to pass any of the unit tests now, since it's just a hack on the existing converters. But I wanted to make a bit of progress on #175. This ran one example network that I was sent, but I've done zero validation.

### DIFF
--- a/converters/keras_v2_layer_converters.py
+++ b/converters/keras_v2_layer_converters.py
@@ -15,9 +15,7 @@ import sys
 from keras_layer_converters_common import activation_map
 import h5py
 
-def set_globals(backend):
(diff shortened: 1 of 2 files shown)


matthewfeickert:
> But it looks like it doesn't run (lint runs for some reason though). I'm a bit better with gitlab CI than github actions, does this make sense?

@dguest It disables itself if there is a lack of activity

image

If you go to https://github.com/lwtnn/lwtnn/actions/workflows/ci.yml you can reenable it again.

matthewfeickert:
Now that you've enabled it again, the next push to this branch will trigger the CI.

PR #185, opened 2026-01-22
ci: Test against non-EOL Pythons

matthewfeickert:
* Python 3.9 went EOL on 2025-10-31, so test from Python 3.10 to Python 3.14.
   - Also, the `actions/setup-python` GHA won't set up older Pythons anymore as GitHub has dropped those.
* setuptools and wheel are only needed if building is happening and the build backend is setuptools. Neither is happening in the tests, so these can be removed.


matthewfeickert:
@dguest This is ready for review. CI is passing on my fork.

PR #194, opened 2026-04-09
feat: Upgrade to Eigen v5

### LINKED ISSUE #192: Add support for eigen v5
[Eigen v5](https://gitlab.com/libeigen/eigen/-/releases/5.0.0) is released and the conda-forge migrations have started over at https://github.com/conda-forge/lwtnn-feedstock/pull/9. It would be great to adopt this, but `lwtnn` currently only supports `eigen` `v3`.

matthewfeickert:
* Resolves #192 

---

* Update required C++ standard to C++14.
* Add cassert includes to header files.
   - `cassert` was provided transitively through Eigen v3 headers, but in Eigen v5 needs to be explicitly included.
* Install Eigen v5 if not provided.
* Use Eigen v5's package configuration file through `NO_MODULE`.
   - Use 'find_package( Eigen3 REQUIRED NO_MODULE )', where NO_MODULE forces Config mode which searches for the specified package configuration file. Eigen v5 distributes this file.
   - c.f. https://libeigen.gitlab.io/eigen/docs-5.0/TopicCMakeGuide.html
* Remove cmake/FindEigen3.cmake as Config mode explicitly ignores it and so it will no longer be used.
* Use `target_link_libraries( ... PUBLIC Eigen3::Eigen )` to use the correct include directories.


matthewfeickert:
@dguest Can you please enable the CI workflow to run again under https://github.com/lwtnn/lwtnn/actions/workflows/ci.yml ?

PR #195, opened 2026-05-26
Add ONNX converter for dense networks

trahxam:
Adds `converters/onnx2json.py` mirroring `kerasfunc2json.py` to convert ONNX graphs into lwtnn JSON.

## Supported ops
- `Gemm` and fused `MatMul`+`Add` as dense layers
- `BatchNormalization`, `Concat`, graph-level `Add` merge
- SiLU/Swish via the `Mul(x, Sigmoid(x))` pattern
- Activations: Relu, Sigmoid, Tanh, Softmax, HardSigmoid, LeakyRelu, Elu
- Skip ops dropped: Identity, Cast, Dropout, Flatten, Reshape, Squeeze, Unsqueeze

## Tests added
- `test-onnx-dense.sh` - Gemm + activations
- `test-onnx-batchnorm.sh` - BatchNorm + Concat + Add merge
- `test-onnx-activations.sh` - MatMul+Add fusion, alpha activations, SiLU, full skip-op chain
- `test-torch-onnx-roundtrip.sh` - end-to-end torch -> ONNX -> lwtnn, validated against torch forward pass

CI gains a `torch-roundtrip` job that installs CPU torch on a single Python version.

Tested to work with ATLAS pixel splitting networks, albeit outside athena.


.github/workflows/ci.yml
```diff
@@ -87,6 +87,46 @@ jobs:
...
+        cmake
+        libhdf5-serial-dev
+        libboost-dev
+        libeigen3-dev
+    - name: Install Python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install --requirement tests/requirements.txt
+        python -m pip install --index-url https://download.pytorch.org/whl/cpu torch
+        python -m pip install onnxscript
```
dguest: @matthewfeickert usually had opinions here, but my gut feeling would be that we should:
- make another requirements file
- pin the pytorch and onnxscript dependencies in some way

I don't know if pytorch and onnxscript make any promises about forward compatibility. At least pinning the major (maybe minor, depending on how strict they are) version(s) might save us some headache when the versions you're pointing to are updated.

converters/onnx_op_converters.py
```diff
@@ -0,0 +1,164 @@
...
+        if a.type == 1:
+            return a.f
+        if a.type == 2:
+            return a.i
+        if a.type == 3:
+            return a.s.decode()
+        if a.type == 6:
+            return list(a.floats)
+        if a.type == 7:
+            return list(a.ints)
```
dguest: Can you use the enums these map to, e.g. `AttributeProto.FLOAT` etc?

dguest (commented):
This is great. Not that the MR needs it but can you comment here on what you're using it for?

